PR #47, opened 2026-01-16
[FEAT: 035] Webhook Server

### LINKED ISSUE #35: Webhook Server for Zot
The solar-discovery-worker needs the ability to receive Webhook events for changes in a OCI registry.

Initially we want to start with support for https://zotregistry.dev/ (also see https://zotregistry.dev/v2.1.12/articles/events/?h=event).

Zot uses cloud events: https://github.com/project-zot/zot/blob/708adf63d4e0d5875200cf869df9acbb3420b58a/pkg/extensions/events/events.go#L69

Acceptance Criteria:
* Webhook Server has options to configure required credentials, basic auth or bearer token
* The callback for different registry "flavors" should be active at the same time, but nested, e.g. `${base_url}/webhook/zot`.
* For the received event a `RepositoryEvent` should be published to a go channel
* Behaviour should be tested by embedding/running zot in tests with event configuration and verifying outputted `RepositoryEvent`s

tmeisel:
Closes #35

### DIFF
--- a/pkg/controller/discovery_controller.go
+++ b/pkg/controller/discovery_controller.go
@@ -36,7 +36,7 @@ type DiscoveryReconciler struct {
 	WorkerArgs    []string
 }
 
-//nolint:lll
+// nolint:lll
 // +kubebuilder:rbac:groups=solar.opendefense.cloud,resources=discoveries,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=solar.opendefense.cloud,resources=discoveries/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=solar.opendefense.cloud,resources=discoveries/finalizers,verbs=update
--- a/pkg/discovery/registry.go
+++ b/pkg/discovery/registry.go
@@ -28,7 +28,7 @@ type Registry struct {
 	// Webhook is optional and provides information about requests coming in from the registry
 	Webhook *Webhook `yaml:"webhook"`
 	// ScanInterval specifies how often a full scan will be triggered
-	ScanInterval time.Duration `yaml:"scanInterval" default:"1d"`
+	ScanInterval time.Duration `yaml:"scanInterval" default:"24h"`
 }
 
 func (r Registry) GetURL() string {


trevex:
I added some tests. Embedding zot is a pain, we should work with the container stuff in our integration test and directly ingest the cloud events.

This branch should be rebased and then the copyright headers should be checked again. 

As we only use a single registry, we can drop the registry from the event and just assume that every pipeline step was provided with the regsitry information

tmeisel:
> I added some tests. Embedding zot is a pain, we should work with the container stuff in our integration test and directly ingest the cloud events.

Great, I was struggling with Gingko but didn't want to jump the gun and use testify instead.

> This branch should be rebased and then the copyright headers should be checked again.

Right, I'll do some cleanup in the morning. We were discussing the options to include the changes of your PR but wanted to consult you first. We can bring it up again during the daily.

> As we only use a single registry, we can drop the registry from the event and just assume that every pipeline step was provided with the regsitry information

Right, that's why I added the `Name` to the registry config. Instead of passing around address and auth info with each and every event, we should have a central registry provider, fed by the config, and just pass the name (aka. ID) of the registry to the consumer.

client-go/openapi/api_violations.report
```diff
@@ -40,3 +40,4 @@ API rule violation: names_match,k8s.io/apimachinery/pkg/apis/meta/v1,StatusCause
 API rule violation: names_match,k8s.io/apimachinery/pkg/apis/meta/v1,Time,Time
 API rule violation: names_match,k8s.io/apimachinery/pkg/runtime,Unknown,ContentEncoding
 API rule violation: names_match,k8s.io/apimachinery/pkg/runtime,Unknown,ContentType
+API rule violation: streaming_list_type_json_tags,go.opendefense.cloud/solar/api/solar/v1alpha1,DiscoveryList,ListMeta
```
trevex: Hmm, this is a bit of an issue...

Not quite sure which linter lead to the `omitempty` to `omitzero` change, but this create the API rule violation:
https://github.com/kubernetes/kube-openapi/blob/a19766b6e2d4/pkg/generators/rules/list_type_streaming_tags.go#L69

I double-checked cert-manager as it was the linting reference and `omitempty` is still used there as well:
https://github.com/cert-manager/cert-manager/blob/master/pkg/apis/certmanager/v1/types_certificate.go#L73

cmd/solar-discovery-worker/main.go
```diff
@@ -0,0 +1,160 @@
...
+	eventsChan := make(chan discovery.RepositoryEvent)
+	errChan := make(chan discovery.ErrorEvent)
+
+	errGroup, ctx := errgroup.WithContext(ctx)
+
+	httpRouter := webhook.NewWebhookRouter(eventsChan)
+	httpRouter.WithLogger(log)
+
+	for _, registry := range registries.GetAll() {
+		if err := httpRouter.RegisterPath(registry); err != nil {
```
trevex: Before this call we should make sure that `registry.Webhook.Path` is set and if not, we should not add it to `httpRouter`.

pkg/controller/discovery_controller.go
```diff
@@ -37,12 +36,13 @@ type DiscoveryReconciler struct {
 	WorkerArgs    []string
 }
 
-//+kubebuilder:rbac:groups=solar.opendefense.cloud,resources=discoveries,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=solar.opendefense.cloud,resources=discoveries/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=solar.opendefense.cloud,resources=discoveries/finalizers,verbs=update
-//+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+//nolint:lll
```
trevex: Argh, issue with line length in comments for `controller-gen`. I do believe it is relevant to not have a whitespace though, e.g. `//+kubebuilder without whitespace in the beginning`

trevex (approved):
lgtm

PR #87, opened 2026-01-27
add golangci.yaml with stricter configuration

trevex:
currently the linter uses defaults, but we should probably enable additional linters. I propose to adopt the cert-manager linting config.


trevex:
Will be included via #47

PR #90, opened 2026-01-30
Feature/hydrated target renderer

### LINKED ISSUE #69: Implement target rendering job
Analaog to #68 we need to implement the rendering for hydrated targets.

Acceptance Criteria:
* Based on #68 add another render function
* The output should be close to the poc: https://github.com/opendefensecloud/solar-rendering-poc/tree/main/charts/hydrated-target
* Add tests for the rendering routine

olzemal:
Fixes #69

### DIFF
--- a/pkg/renderer/render_hydrated_target.go
+++ b/pkg/renderer/render_hydrated_target.go
@@ -5,27 +5,19 @@ package renderer
 
 import (
 	"embed"
-	"encoding/json"
 )
 
 //go:embed template/hydrated-target/*
 var hydratedTargetFS embed.FS
 
 type HydratedTargetInput struct {
-	Profiles map[string]Profile `json:"profiles"`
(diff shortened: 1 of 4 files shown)


pkg/renderer/render_hydrated_target.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"embed"
+	"encoding/json"
+)
+
+//go:embed template/hydrated-target/*
+var hydratedTargetFS embed.FS
+
+type HydratedTargetInput struct {
+	Profiles map[string]Profile `json:"profiles"`
```
trevex: Should probably be Releases? (also needs updating in `release.yaml`) Alternatively you can leave Profiles and add Releases as well, as we will skip profiles in the implementation of hydrated targets for now.

trevex:
Some how missed that last night, you are right :)

PR #121, opened 2026-02-05
Implement OCM "qualifier" pipeline step

### LINKED ISSUE #37: Implement OCM "qualifier" pipeline step
The "qualifier" is a part of our proposed discovery pipeline, which receives the `RepositoryEvent` and outputs concrete `ComponentVersionEvent`s.

AC
* Find a good name for this component
* Finds all components and versions for a repository (and optionally tag, if specified)
* Outputs `ComponentVersionEvent`
* Tests should thoroughly test this behaviour
* Throttle to not overwhelm the registry

jastBytes:
Enhance the discovery controller by add missing Kubernetes client and implementing a qualifier pipeline step that includes rate limiting. Additionally, modify the component version lookup to avoid listing all versions when a specific version is provided.

Closes #37.

### DIFF
--- a/cmd/solar-discovery-worker/main.go
+++ b/cmd/solar-discovery-worker/main.go
@@ -18,8 +18,6 @@ import (
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
-	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


jastBytes:
Yeah wasn't sure where the interaction with k8s will happen. if there will be another component that's fine with me. removed the client for now, wasn't used anyway.

PR #122, opened 2026-02-05
ComponentHandler with support for Helm

### LINKED ISSUE #38: ComponentHandler with support for Helm
For each `ComponentVersionEvent` we need to retrieve additional information to complete the description of the ComponentVersion in the API.

A component can be either of type `helm` or `kro`.

For helm we need to:
* Find out if component is classified as helm (check if only single helm chart contained, ...)
* Retrieve all relevant information to create bootstrap manifests for the primary helm chart
* Retrieve default values and schema (for Helm3, see https://helm.sh/docs/topics/charts/#schema-files)
* Test behaviour

jastBytes:
Introduce Helm support in the component handler by refactoring the handler registration process and implementing logic to classify and process Helm components.

Closes #38.


trevex (commented):
In relation to my comment in #121 : Does it make sense that the handler, e.g. helm, returns or sends events with the k8s api types Component and ComponentVersion, so that the writer can than check and patch resources?

jastBytes:
Yes, sounds about right thing to do. Will create a new event type for this and publish them as part of the handler code.

trevex (approved):

PR #153, opened 2026-02-12
Check if ComponentVersion needs updates

### LINKED ISSUE #40: Check if ComponentVersion needs updates
The `ComponentHandler` (#38, #39) is failry expensive and we should avoid doing the work if it is not required.

So we should add a pipeline step, that simply takes an `ComponentVersionEvent` checks if the API server is up to date and only forward the `ComponentVersionEvent` if the version is not registered.

AC:
* Check API server if ComponentVersion exists
* Forward only if update
* Test behavior with apiserver testenv

jastBytes:
Closes #40.

Sidequest:

Unifies and simplifies event-driven logic by introducing a generic Runner structure for handling event processing, logging, rate limiting, and backoff. Reduces boilerplate and improves code maintainability by refactoring qualifier and related components to use this reusable mechanism. Enhances test clarity and alignment with new abstractions.

### DIFF
--- a/pkg/discovery/events.go
+++ b/pkg/discovery/events.go
@@ -67,8 +67,6 @@ type WriteAPIResourceEvent struct {
 type ErrorEvent struct {
 	// Error is when an error occurred.
 	Error error
-	// Message is an optional message providing additional context about the error.
-	Message string
 	// Timestamp is the timestamp when the event was created.
 	Timestamp time.Time
 }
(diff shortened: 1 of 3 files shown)


pkg/discovery/helpers.go
```diff
@@ -23,3 +25,41 @@ func SplitRepository(repo string) (string, string, error) {
...
+
+// SanitizeName cleans a string to be a valid K8s resource name.
+// It ensures:
+// 1. Max 63 characters
+// 2. Lowercase alphanumeric or '-'
+// 3. Starts and ends with alphanumeric
+func SanitizeName(input string) string {
+	name := strings.ToLower(input)
+
+	reg := regexp.MustCompile("[^a-z0-9]+")
```
tmeisel: Die Zeile würde ich auslagern, muss ja nicht immer wieder compilen.

jastBytes: You're right. Change incoming!

tmeisel (approved):

Issue #123, opened 2026-02-05
GetEventRecorderFor is deprecated


trevex:
See cmd/solar-controller-manager/main FIXME comments and also in ARC cmd/arc-controller-manager/main

PR #156, opened 2026-02-12
Replace deprecated GetEventRecorderFor

yocaba:
The `GetEventRecorderFor` function was deprecated in the latest K8s update. Per the documentation:

```
// Deprecated: this uses the old events API and will be removed in a future release. Please use GetEventRecorder instead.
```

I have replaced `GetEventRecorderFor` with `GetEventRecorder`. The latter returns an `EventRecorder` object from a different package, which has a slightly different signature for `Eventf`. I have adjusted the `Eventf` calls, passing a second resource object where applicable or `nil` and a new action string (alongside to reason).

Closes https://github.com/opendefensecloud/solution-arsenal/issues/123.

### DIFF
--- a/cmd/solar-controller-manager/main.go
+++ b/cmd/solar-controller-manager/main.go
@@ -187,11 +187,9 @@ func main() {
 
 	// Register controllers
 	if err := (&controller.DiscoveryReconciler{
-		Client: mgr.GetClient(),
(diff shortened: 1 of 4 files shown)


jastBytes (approved):
/lgtm

PR #186, opened 2026-02-20
Update to helm v4

### LINKED ISSUE #144: Update helm v3 to v4
We still use v4 in our push chart implementation of the renderer

trevex:
Closes #144


trevex:
Several fields are not supported by Helm v4 anymore, e.g. you can not create an insecure TLS client.
I removed all of them from the API.

olzemal (approved):

PR #194, opened 2026-02-23
Add version checking for installing make targets

### LINKED ISSUE #190: Makefile new versions not updated
When the version of a binary is bumped, the Makefile seems to not properly handle the version check and is not correctly updated.

@olzemal fixed this, but somehow this is not fixed anymore.

olzemal:
Fixes #190


yocaba (approved):
lgtm

Issue #184, opened 2026-02-20
Reference Credentials from a Secret


olzemal:
**Is your feature request related to a problem? Please describe.**
Currently we expect Credentials to be directly in a RenderTask. This should be a Reference to a Secret. In order to comply with kubernetes standards.

**Describe the solution you'd like**
The Credentials Fields in Public API Resources are removed and replaced by a secret reference.

The RenderTask Controller resolves the secret reference and makes the contents available to the renderer job.

**Describe alternatives you've considered**
We considered making the renderer poll directly from a kubernetes resource and resolve the secret reference itself, but decided against it in order to remain the standalone capabilities in case of a disaster recovery.

**Additional context**
https://github.com/opendefensecloud/solution-arsenal/blob/29e24884d4dc8f2c1e8be684eb1a837ab71300c6/api/solar/renderer_types.go#L74

This might also be applied to the discovery-worker

PR #195, opened 2026-02-23
Reference Credentials from Secret

olzemal:
## Issue

- [x] #184 

## Description

This PR replaces the Password in RenderTask Spec with a SecretRef.

Implicitly RenderTask Controller needs logic to resolve the secret and create jobs with the needed credentials.

The renderer CLI also needs to be adjusted in order for the secrets to be passed by flags / environment vars.

## Checklist
- [x] I've added tests that cover the changes
- [x] I've updated the documentation to reflect the new behavior


yocaba (commented):
I might miss pieces of the overall picture. If it's more effective, let's talk to address some of my comments

PR #211, opened 2026-03-02
Add nolint for discovery registry password

olzemal:
as discussed in #184


pkg/discovery/registry.go
```diff
@@ -15,6 +15,12 @@ type RegistryCredentials struct {
 	// Username is the username used to authenticate with the registry
 	Username string `yaml:"username"`
 	// Password is the password used to authenticate with the registry
+	// If the discovery worker is run as a standalone binary users would have to
+	// write the registry password into the configuration file. This use-case is
+	// not supported for now since the configuration gets built by the
```
yocaba: At least it isn't supported as a production use case. For testing purposes we wanted to support it so that one can run the discovery without a K8s cluster?

Issue #101, opened 2026-02-02
Test target registration and rendering


trevex:
1. Create/register a Target
2. Expect HydratedTarget to be created eventually
3. Expect Helm Chart to be rendered to OCI eventually

PR #273, opened 2026-03-12
Target e2e test

olzemal:
Relates to #101

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -11,13 +11,11 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"fmt"
-	"net"
 	"net/http"
 	"os/exec"
 	"path/filepath"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/rand"
 	"oras.land/oras-go/v2/registry"
 	"oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -159,8 +285,65 @@ var _ = Describe("solar", Ordered, func() {
...
+	cmd := exec.Command(kubectlBinary, finalargs...)
+	setCmdContext(cmd)
+	cmd.Stdout = GinkgoWriter
+	cmd.Stderr = GinkgoWriter
+
+	err := cmd.Start()
+	Expect(err).NotTo(HaveOccurred())
+
+	Eventually(func() error {
+		conn, err := net.DialTimeout("tcp", fmt.Sprintf("localhost:%d", localport), time.Second)
```
yocaba: Minor: We better should use the IP to not require a working hostname resolving.

hack/dev-cluster.sh
```diff
@@ -71,7 +71,7 @@ setup_trust_manager() {
 
 setup_zot_certs() {
     echo -e "\nSETTING UP CERT FOR ZOTs:\n"
-    $KUBECTL create namespace zot
+    $KUBECTL get namespace zot 2>/dev/null || $KUBECTL create namespace zot
```
yocaba: Why don't we care that the namespace already exists?

yocaba (changes requested):
A couple of minor things and clarification questions. Overall nice improvement to avoid having two clusters with different setup for dev and e2e.

Issue #275, opened 2026-03-12
Failure in discovery's APIWriter for update event


yocaba:
Setup the test cluster with `make dev-cluster`. Transfer the helm demo OCM package as done here but to the Zot registry that is expected to be configured with webhook support:

https://github.com/opendefensecloud/solution-arsenal/blob/0855ff0596bb9e69d2854abf72398e48b55afe78/Makefile#L226

Create a Discovery resource like https://github.com/opendefensecloud/solution-arsenal/blob/main/test/fixtures/e2e/discovery.yaml.

Following error occurs:

```
[...]

PR #280, opened 2026-03-13
Add component spec to WriteAPIResourceEvents

yocaba:
Closes https://github.com/opendefensecloud/solution-arsenal/issues/275.

We missed to add the Component spec to the WriteAPIResourceEvents in the helm handler so that the APIWriter complained about missing data.


olzemal (approved):
lgtm

PR #317, opened 2026-03-20
Add ADR for moving RenderTask jobs to dedicated namespace

### LINKED ISSUE #198: Run RenderTask jobs in dedicated namespace
Currently the jobs are scheduled in the namespace of the RenderTask object.

This might be undesirable:
* complicates secret handling (user should not have access to "render oci registry")
* is actually something done for the user, but owned by the operator

Also take comments below into consideration.

AC:
* Evaluate and document different options
* Decide (with the team) how to proceed and document decision
* Create/refine follow up tickets as necessary

olzemal:
Fixes #198

## Summary by CodeRabbit

* **Documentation**
  * Added an architectural decision record: RenderTask will become a cluster-scoped resource and its execution Jobs will run in the controller namespace. The record summarizes evaluated alternatives, lists rejected options, and links a follow-up issue for implementation.


dermorz (approved):
lgtm

PR #327, opened 2026-03-25
[FEAT: 212] Webhook refactor

tmeisel:


olzemal:
fixed in https://github.com/opendefensecloud/solution-arsenal/pull/346

Issue #391, opened 2026-04-09
Create a minimal in-cluster next.js app for solar


rebEllieous:
This spike aims to implement a minimal proof-of-concept (PoC) to validate this approach within the SOLAR architecture.
Goals

The primary objective is to verify that a frontend-only application can securely and efficiently communicate with the K8s API server from within the cluster.

### Key Areas of Investigation

- Authentication & Authorization: How does the SPA handle credentials? (service account? OIDC-based auth?).
- Client Library: Evaluating the use of the official @kubernetes/client-node vs. generated openapi client.

### Acceptance Criteria

- A working PoC showing successful data retrieval from the K8s API.
- Documentation on how the SPA handles authentication.
- A brief summary of trade-offs encountered during implementation (specifically regarding security and CORS).

rebEllieous:
spike is outdated, ui prototype in https://github.com/opendefensecloud/solution-arsenal/pull/397

PR #399, opened 2026-04-13
feat: poc minimal solar-ui dev setup

rebEllieous:
draft for #391


test/fixtures/helmdemo-ctf/artifact-index.json
olzemal, line 1: Shouldn't this be in `.gitignore`?

PR #400, opened 2026-04-14
feat: add targetNamespace to Release

### LINKED ISSUE #239: Add namespace/targetNamespace to Release
The release is missing the namespace to deploy to!

olzemal:
Fixes #239 

Successor of #299

- Add targetNamespace field to Releases
- Render targetNamespace in HelmRelease
- Generate RenderTasks with targetNamespace in RendererConfig
- Adjust end to end test to both verify targetNamespace gets set as well as look for deployments in the correct namespaces.

## Summary by CodeRabbit

* **New Features**
  * Optional targetNamespace added to Release/ReleaseConfig and client apply builders; renderer and controller propagate it and set createNamespace only when provided.
  * Generated manifests include stable release/component labels plus annotations with original names for traceability.

* **Documentation**
  * API reference and OpenAPI schemas updated to document targetNamespace.

* **Tests**
  * Unit, renderer and e2e tests added/updated to validate targetNamespace behavior, rendering, deployment and cleanup.

* **Style**
  * Minor internal grouping changes for compile-time assertions.

### DIFF
--- a/pkg/renderer/template/bootstrap/templates/release.yaml
+++ b/pkg/renderer/template/bootstrap/templates/release.yaml
@@ -7,12 +7,22 @@
   {{- $name = printf "%s-%s" ($name | trunc 42) $sha }}
 {{- end }}
 
+{{- $releaseLabel := $k }}
+{{- if gt (len $releaseLabel) 63 }}
+  {{- $sha := $releaseLabel | sha256sum | trunc 10 }}
+  {{- $releaseLabel = printf "%s-%s" ($releaseLabel | trunc 52) $sha }}
+{{- end }}
+
 ---
 apiVersion: source.toolkit.fluxcd.io/v1
 kind: OCIRepository
 metadata:
   name: {{ $name }}
   namespace: {{ $.Release.Namespace }}
+  labels:
+    solar.opendefense.cloud/release: {{ $releaseLabel }}
+  annotations:
+    solar.opendefense.cloud/release: {{ $k }}
 spec:
   interval: 10m
   url: oci://{{ $v.repository }}
@@ -33,6 +43,10 @@ kind: HelmRelease
 metadata:
   name: {{ $name }}
   namespace: {{ $.Release.Namespace }}
+  labels:
+    solar.opendefense.cloud/release: {{ $releaseLabel }}
+  annotations:
+    solar.opendefense.cloud/release: {{ $k }}
 spec:
   interval: 10m
   chartRef:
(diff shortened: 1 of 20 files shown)


olzemal:
@coderabbitai review

test/e2e/e2e_test.go
```diff
@@ -455,27 +470,107 @@ var _ = Describe("solar", Ordered, func() {
 				}
 			}).Should(Succeed())
 
+			By("verifying HelmRelease has targetNamespace set")
```
yocaba: Should the E2E test verify that the target namespace is internally 'set,' or should it focus on the final outcome—confirming the app is actually deployed in the correct namespace? Given that we already have unit tests covering the configuration logic?

pkg/renderer/template/bootstrap/templates/release.yaml
```diff
@@ -7,12 +7,22 @@
...
+{{- end }}
+
 ---
 apiVersion: source.toolkit.fluxcd.io/v1
 kind: OCIRepository
 metadata:
   name: {{ $name }}
   namespace: {{ $.Release.Namespace }}
+  labels:
+    solar.opendefense.cloud/release: {{ $releaseLabel }}
```
yocaba, line 23: Where does this label come from? How is it related to the target namespace change?

test/e2e/e2e_test.go
olzemal: It does both. Verifying the deployments checks this implicitly https://github.com/opendefensecloud/solution-arsenal/blob/de183f1ac248b373eb15e1307c8fce5cf10998f4/test/e2e/e2e_test.go#L569

pkg/renderer/template/bootstrap/templates/release.yaml
olzemal: I added the label in order to make finding the related helmreleases more easily since the names are appended and limited to 63 chars resulting in not really readable resource names. I made use of this in the e2e test, maybe it could also be helpful for users. For example here:

https://github.com/opendefensecloud/solution-arsenal/blob/de183f1ac248b373eb15e1307c8fce5cf10998f4/test/e2e/e2e_test.go#L476

test/e2e/e2e_test.go
yocaba: Yes, it does already both, but what I meant, to keep the e2e tests as simple as possible: Can we skip checking the spec and leave that for the unit tests and only verify the namespace (implicitly) when verifying the deployments?

olzemal: Hmm, i guess that could work. I dont know how to feel about implicit trust between tests but maybe that's just me 😅

olzemal:
@coderabbitai review

yocaba (approved):
lgtm

Issue #246, opened 2026-03-06
Implement release resolver


olzemal:
Multipe releases can be bound to the same target and have conflicts via uniqueName or anti affinities. We need to resolve those as part of the target reconcilation.

### Acceptance Criteria

- [ ] Target controller checks unique names: if two releases with the same unique name for the same target, only the higher priority release is deployed
- [ ] Target controller evaluates anti-affinity rules: if a release declares anti-affinity against another release already bound to the target, the release is not rendered
- [ ] Priority ordering is deterministic (tiebreaker: alphabetical profile name)
- [ ] Status conditions on Target reflect any skipped bindings due to priority or anti-affinity
- [ ] Tests cover priority resolution, anti-affinity conflicts, and edge cases

### Dependencies

- #242 Add unique name fields to Release API
- #243 Add priority field to Profile API
- #244 Add anti-affinity fields to Release API

PR #444, opened 2026-04-22
feat: add uniqueName to ReleaseSpec

### LINKED ISSUE #242: Add Unique Name Fields to Release API
Calculate unique name from Component name and make it available in Release Status.

Add an optional Parameter to the Release Spec that can be defined to override the Unique name.

related to #172

dermorz:
Closes #242.

Adds a required `uniqueName` field to `ReleaseSpec`. The field serves as a logical identifier for a component deployment and will be used by the Profile controller (#246) to enforce that a given component (e.g. Kyverno) is deployed only once per target cluster, even when multiple Profiles match the same target.

The field is required rather than optional — there is no implicit fallback, and the ADR treats the unique name as fundamental to every Release.

## Summary by CodeRabbit

* **New Features**
  * Added a required ReleaseSpec field `uniqueName` to uniquely identify releases and avoid duplicate component deployments when multiple profiles match the same cluster.

* **Validation**
  * Creation requires non-empty `spec.uniqueName`; updates reject empty or changed `uniqueName` (immutable).

* **Documentation**
  * API reference updated to document the new `uniqueName` field.

* **Tests & Fixtures**
  * Tests and example manifests updated to include and validate `uniqueName`.

### DIFF
--- a/api/solar/release_rest.go
+++ b/api/solar/release_rest.go
@@ -85,7 +85,13 @@ func (o *Release) Validate(ctx context.Context) field.ErrorList {
 }
 
 func (o *Release) ValidateUpdate(ctx context.Context, old runtime.Object) field.ErrorList {
-	return validateRelease(o)
+	errors := validateRelease(o)
+	or := old.(*Release)
+	if o.Spec.UniqueName != or.Spec.UniqueName {
+		errors = append(errors, field.Forbidden(field.NewPath("spec").Child("uniqueName"), "uniqueName is immutable"))
+	}
+
+	return errors
 }
 
 func validateRelease(o *Release) field.ErrorList {
--- a/api/solar/release_rest.go
+++ b/api/solar/release_rest.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/duration"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 var (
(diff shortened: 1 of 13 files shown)


api/solar/release_rest.go
```diff
@@ -54,3 +57,20 @@ func (o *Release) PrepareForUpdate(ctx context.Context, old runtime.Object) {
...
+}
+
+func (o *Release) ValidateUpdate(ctx context.Context, old runtime.Object) field.ErrorList {
+	return validateRelease(o)
+}
+
+func validateRelease(o *Release) field.ErrorList {
+	errors := field.ErrorList{}
+	if o.Spec.UniqueName == "" {
+		errors = append(errors, field.Required(field.NewPath("spec").Child("uniqueName"), "uniqueName must not be empty"))
```
trevex, line 100: Is uniqueName truly required?

For some releases certainly, we want to allow in-place replacement. The background was the user story around kyverno: Two distinct releases of Kyverno. One for Kyverno V1 another for Kyverno V2. One matching older cluster, the other newer K8s version. We want in-place upgrades so we need a chain of deterministic HelmRelease names.

I am not certain this is true for every release though.

dermorz: As we discussed offline, we keep it required for now. Later on we can think about improving user experience by making it optional with a fallback to `Component.Name`.

api/solar/release_rest.go
```diff
@@ -76,3 +79,26 @@ func (o *Release) ConvertToTable(ctx context.Context, tableOptions runtime.Objec
...
+func (o *Release) Validate(ctx context.Context) field.ErrorList {
+	return validateRelease(o)
+}
+
+func (o *Release) ValidateUpdate(ctx context.Context, old runtime.Object) field.ErrorList {
+	errors := validateRelease(o)
+	or := old.(*Release)
+	if o.Spec.UniqueName != or.Spec.UniqueName {
+		errors = append(errors, field.Forbidden(field.NewPath("spec").Child("uniqueName"), "uniqueName is immutable"))
+	}
```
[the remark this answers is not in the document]
dermorz: We are still in early development and there are no long living "live systems" running SolAr. Only short-lived dev- or test-setups.

trevex (approved):

Issue #189, opened 2026-02-23
Improve code quality of discovery


trevex:
I recently ran a ML based code quality analysis over a PR to test it.
The output was not relevant to the changes of the PR, but we should probably look at them regardless:

# Code Quality Analysis: `feat/discovery-pipeline-webhook` Branch

## Overview
Analysis of code quality in the `feat/discovery-pipeline-webhook` branch compared to `main`.

---

## ✅ Good Practices Found

1. **Proper error handling in webhook server** - Errors from Serve() are properly published to error channel
2. **Good test coverage** - Webhook tests cover success and failure cases
3. **Clean formatting** - All files properly formatted with gofmt
4. **Type safety** - No go vet warnings
5. **Proper context usage** - Contexts with timeouts used appropriately
6. **Graceful shutdown** - Server implements proper Stop() with timeout

---

### 1. **Unused Generic Type Parameters** (Minor)
**File**: `pkg/discovery/pipeline/pipeline.go` (Lines 130-151)

The `Option` functions have unused generic type parameters:
```go
func WithQualifierProcessor[InputEvent any, OutputEvent any](proc discovery.Processor[...]) Option {
    // InputEvent and OutputEvent are never used
}
```

**Impact**: Low - misleading but doesn't break functionality

**Recommendation**: Either use the parameters or remove them

---

### 2. **Missing Error in Stop() Return** (Moderate)
**File**: `pkg/discovery/webhook/server.go` (Lines 60-66)

The `Stop()` method doesn't return errors even though shutdown can fail:
```go
func (s *WebhookServer) Stop(ctx context.Context) {
    if err := s.server.Shutdown(shutdownCtx); err != nil {
        s.log.Error(err, "Warning: server shutdown error")  // Just logs, doesn't return
    }
}
```

**Impact**: Moderate - Calling code can't know if shutdown failed

**Recommendation**: Return `error` to allow proper error handling:
```go
func (s *WebhookServer) Stop(ctx context.Context) error {
    // ... return err if shutdown fails
}
```

**Files affected**:
- `pkg/discovery/webhook/server.go` (signature)
- Any callers of `Stop()` method

---

### 3. **Inconsistent Error Handling Pattern** (Moderate)
**File**: `pkg/discovery/scanner/registry_scanner.go`

Changed from `rs.handleError()` to just `rs.logger.Error()` in one location (line 160), but still uses `handleError()` elsewhere:
```go
// Line 160: Direct log
[...]

PR #447, opened 2026-04-22
chore: code quality improvements for discovery #189

rebEllieous:
solves #189 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown/error handling across discovery pipeline and webhook server to capture and return stop errors and surface failures.
  * More precise handling and logging for registry/repository listing errors.

* **Tests**
  * Updated tests to explicitly handle stop errors and to poll for post-stop behavior where appropriate.

* **Chores**
  * Expanded .gitignore to exclude certificate, key, frontend build, and dependency artifacts.


pkg/discovery/scanner/registry_scanner.go
```diff
@@ -153,7 +154,9 @@ func (rs *RegistryScanner) Scan(ctx context.Context, eventsChan chan<- discovery
 		return
 	}
 
-	// Lists all repositories in the registry
+	// Lists all repositories in the registry. Per-repository errors
+	// are only logged, not published to errChan: they are expected
+	// and must not terminate the scan loop.
```
yocaba, line 159: Does the current implementation match the comment? 🤔 If an error occurred that is not related to cancelling, `rs.handleError` is called which sends the error into the error channel.

rebEllieous: the outer err returned from client.Repositories is a fatal one which is propagated up via the errChan, as we cant continue with any meaningful work without a repository listing. the inner err from processRepository is handled in a non-panic-y way, where execution is still resumed if singular repositories fail inside the process loop

yocaba (dismissed):

yocaba (approved):

PR #452, opened 2026-04-24
fix(ci): prevent incorrect helm package selection

### LINKED ISSUE #450: Releasing v0.1.0 failed on helm publish
https://github.com/opendefensecloud/solution-arsenal/actions/runs/24878335762/job/72841010626

yocaba:
Closes #450.

The issue: In the second loop iteration (for the chart solar), the package name was derived using:

```
ls ${CHART_NAME}-*.tgz
```

Because `solar-discovery-0.1.0.tgz` was already present in the workspace from the first iteration, the `ls` command matched both `solar-0.1.0.tgz` and `solar-discovery-0.1.0.tgz`. This resulted in the `PACKAGE` variable containing multiple filenames, causing `helm push` to fail.

## Summary by CodeRabbit

* **Chores**
  * Improved Helm chart publishing workflow with more precise artifact selection and automated cleanup of local package files after deployment.

### DIFF
--- a/.github/workflows/helm-publish.yaml
+++ b/.github/workflows/helm-publish.yaml
@@ -103,7 +103,7 @@ jobs:
               helm package "${chart}"
 
               # Push to OCI registry
-              PACKAGE=$(ls ${CHART_NAME}-*.tgz)
+              PACKAGE=$(ls ${CHART_NAME}-[0-9]*.tgz)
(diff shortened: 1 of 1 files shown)


olzemal (approved):
lgtm, a bit confusing that this is not regex but globs. Maybe using find / grep would make this more obvious but thats just a nitpick and probably overkill :)

PR #471, opened 2026-04-28
feat: add antiAffinity field to ReleaseSpec

### LINKED ISSUE #244: Add Anti-Affinity to Release API
Releases should define anti-affinity rules based on labels.

For example a release of linkerd could define that it is exclusive to istio.

related to https://github.com/opendefensecloud/solution-arsenal/issues/172

dermorz:
## What
Adds `antiAffinity` field to `ReleaseSpec` to define exclusion rules preventing conflicting software from being deployed to the same target.
Closes #244

## Why
Enables releases to declare anti-affinity against other releases based on label selectors. For example, a Linkerd release can declare anti-affinity against Istio by matching on a `service-mesh` category label.

## Testing
Unit tests added in `api/solar/release_rest_test.go` covering:
- nil antiAffinity acceptance
- populated antiAffinity with MatchLabels
- antiAffinity with MatchExpressions
- JSON round-trip for both MatchLabels and MatchExpressions
- omission from JSON when nil

## Notes for reviewers
- API change: Added `AntiAffinity *metav1.LabelSelector` to `ReleaseSpec`
- Codegen updated: deepcopy, conversion, client-go apply config, openapi
- Related: Controller logic for evaluating anti-affinity will be implemented in #246

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added anti-affinity support to Release deployments: specify label selectors to prevent deployment when another matching Release is already bound to the same target.

* **Documentation**
  * Updated API reference documentation to clarify optional fields across multiple resource types.

### DIFF
--- a/api/solar/release_types.go
+++ b/api/solar/release_types.go
@@ -21,6 +21,11 @@ type ReleaseSpec struct {
 	// UniqueName is a logical identifier used to ensure this component is deployed
 	// only once per target cluster when multiple Profiles match the same target.
 	UniqueName string `json:"uniqueName"`
+	// AntiAffinity defines exclusion rules. If another Release matching this
+	// label selector is already bound to the same Target, this Release should
+	// not be deployed there (or a conflict condition should be raised).
+	// +optional
+	AntiAffinity *metav1.LabelSelector `json:"antiAffinity,omitempty"`
 	// Values contains deployment-specific values or configuration for the release.
 	// These values override defaults from the component version and are used during deployment.
 	// +optional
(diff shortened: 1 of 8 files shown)


api/solar/release_rest_test.go
```diff
@@ -80,6 +81,111 @@ var _ = Describe("Release REST", func() {
 		})
 	})
 
+	Describe("AntiAffinity field", func() {
```
yocaba: Just wondering what's the purpose of the tests given that we didn't add custom validation logic. 🤔

dermorz: You are actually not wrong here. We could just rely on the `metav1.LabelSelector` type's own tests + codegen and drop the tests here. Or we keep them as kind of API contract documentation or to avoid regressions?

I am fine either way.

yocaba: I’d prefer then to drop the tests; it wouldn't scale well for the entire API and I think the risk of missing accidental API changes with the current setup is low.

yocaba (approved):

PR #474, opened 2026-04-28
feat: poc implementation resource grants for common use cases

### LINKED ISSUE #314: Spike: Re-evaluate cluster-scoped vs namespaced with ReferenceGrants
During discussion of #174 [ReferenceGrants](https://gateway-api.sigs.k8s.io/api-types/referencegrant/) (from the Gateway API project) were identified as a potential mechanism to enable secure cross-namespace references without requiring cluster-scoped resources.

AC:
* PoC Implementation of ReferenceGrants in SolAr context
* Evaluation if ReferenceGrants enable a pure namespaced approach for all desired use-cases
* Re-evaluate cluster-scoped options based on findings

rebEllieous:
## What
Implements a ResourceGrant CRD enabling three cross-namespace reference patterns: Profile→Target, Target→Registry, and Release→ComponentVersion
Closes #314

## Testing
Integration and e2e Tests were added

## Notes for reviewers
- New CRD: ResourceGrant
- ReleaseSpec gains componentVersionNamespace (optional string). Same-namespace behaviour is unchanged.
- TargetSpec gains renderRegistryNamespace (optional string). Same-namespace behaviour is unchanged.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added ReferenceGrant CRD to enable secure cross-namespace references.
  * Releases can resolve ComponentVersions in other namespaces.
  * ReleaseBindings can target Targets in different namespaces.
  * Targets can reference render Registries in other namespaces.

* **Documentation**
  * API reference updated with cross-namespace fields and ReferenceGrant spec.

* **Tests**
  * Added controller and end-to-end tests validating cross-namespace behaviors and grants.


pkg/controller/target_controller.go
```diff
@@ -154,6 +159,22 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, errLogAndWrap(log, err, "failed to get Registry")
 	}
 
+	// If the registry lives in a different namespace, verify a ReferenceGrant permits it.
```
dermorz: I think we want to do this check before getting the registry (L146-160), so we don't get the object at all if not granted.

pkg/controller/profile_controller.go
```diff
@@ -209,5 +331,73 @@ func (r *ProfileReconciler) mapTargetToProfiles(ctx context.Context, obj client.
...
+		return requests
+	}
+
+	for i := range grantList.Items {
+		grant := &grantList.Items[i]
+		if !grantsTargetResource(grant) {
+			continue
+		}
+		for _, from := range grant.Spec.From {
+			if from.Namespace == target.Namespace {
```
dermorz: Let's also add a filter on `from.Kind`:

```
if from.Kind != "Profile" {
    continue
}
```

There could be multiple Grants, even for other kinds, so let's skip these too.

pkg/controller/profile_controller.go
```diff
@@ -67,18 +145,49 @@ func (r *ProfileReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	for _, target := range targetList.Items {
-		desiredTargets[target.Name] = true
+	allTargets := make([]solarv1alpha1.Target, 0, len(sameNsTargets.Items))
+	allTargets = append(allTargets, sameNsTargets.Items...)
+
+	// Collect cross-namespace targets via ReferenceGrants.
+	// A ReferenceGrant in namespace B listing the profile's namespace in From and
+	// "targets" in To allows this Profile to select Targets from namespace B.
+	grantList := &solarv1alpha1.ReferenceGrantList{}
+	if err := r.List(ctx, grantList); err != nil {
```
dermorz, line 169: There is no namespace filter on this `List()`, so we are getting _all_ `ReferenceGrants` across the whole cluster on every `Profile` reconcile. This might become a performance/resource issue at scale (many namespaces, many profiles).

It's only querying the informer cache, so it's just a memory scan and not an etcd round-trip - and since SolAr's resources live in our own etcd, we also wouldn't impact the host cluster's control plane.

But let's still maybe add a comment here to keep it in mind.

api/solar/v1alpha1/referencegrant_types.go
```diff
@@ -0,0 +1,87 @@
...
+	// +kubebuilder:validation:Required
+	Kind string `json:"kind"`
+}
+
+// ReferenceGrantSpec defines the desired state of a ReferenceGrant.
+type ReferenceGrantSpec struct {
+	// From is the list of resources that are permitted to reference resources in this namespace.
+	// Each entry specifies the group, kind, and namespace of an allowed referencing resource.
+	// +kubebuilder:validation:MinItems=1
+	From []ReferenceGrantFromSubject `json:"from"`
```
dermorz, line 48: There are new entries in [client-go/openapi/api_violations.report](https://github.com/opendefensecloud/solution-arsenal/pull/474/changes#diff-4d329f0a312e1db231da6cb94471376b4a8fb2ec6a87850b059be311e21feab8):
```diff 
+ API rule violation: list_type_missing,go.opendefense.cloud/solar/api/solar/v1alpha1,ReferenceGrantSpec,From
+ API rule violation: list_type_missing,go.opendefense.cloud/solar/api/solar/v1alpha1,ReferenceGrantSpec,To
```

To fix this for `From` and `To`, add `// +listType=atomic` to both fields (in both `referencegrant_types.go` files) like here: https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/referencegrant_types.go#L74-L91

Then run `make codegen` again.

---

`+listType` tells Server-Side Apply (SSA) how to merge list fields. Without it, the API machinery falls back to an unspecified default, and the API
  linter flags it as a violation. The three options are `atomic` (replace the whole list), `set` (deduplicated values), and `map` (merge by key field).

We want `atomic` here like APIGateway, because a grant is owned by a single manager.

dermorz (changes requested):
Claude helped me a bit. I didn't even know about the `api_violations.report` - where you didn't introduce the first entries, so we might want to fix the others too.

dermorz (approved):
lgtm 👍

PR #494, opened 2026-05-05
feat(controller): implement release resolver in Target controller (#246)

dermorz:
## What

Adds a deterministic release resolver to the Target controller that runs after binding collection and before RenderTask creation. Implements uniqueName-based deduplication (priority wins; bindingKey tiebreaker) and bidirectional anti-affinity enforcement.

Also makes `spec.uniqueName` optional on Release: when not set, the effective unique name falls back to the parent Component name from the referenced ComponentVersion, written to `status.effectiveUniqueName` for operator visibility.

## Why
Without conflict resolution, two Profiles can bind the same Target to two versions of the same component (e.g., kyverno v3.2 and v3.3), resulting in duplicate RenderTasks and unpredictable deployments. The resolver enforces that only one Release per logical component reaches the render stage, with the selection governed by explicit priority and anti-affinity rules.

`UniqueName` being optional reduces friction for simple single-profile setups — the system derives a sensible default automatically and exposes it in `status.effectiveUniqueName` so operators can verify without reading docs.

## Testing
- **Unit tests** (`resolveReleaseConflicts`): priority selection, bindingKey tiebreaker, bidirectional anti-affinity, invalid selector, component-name fallback for empty UniqueName.
- **Integration tests** (envtest): higher-priority wins on conflict, alphabetical tiebreaker, anti-affinity blocking, NoConflicts path, optional UniqueName + status.effectiveUniqueName fallback.

Adding e2e scenarios for the resolver was considered but skipped: the resolver is a pure in-process filter with no external I/O. Setting up multiple OCI-registered components, Profiles, and Releases in a live Kind cluster just to test a sort/filter function would add significant fixture overhead for no additional coverage beyond what the envtest integration tests already provide.

## Notes for reviewers
**New condition:** `ReleasesResolved` on Target (`NoConflicts` / `Resolved` with per-binding filter messages / `AllReleaseBindingsFiltered` on ReleasesRendered when all are rejected).

**API change — Release.Spec.UniqueName is now optional:** Previously required and immutable; still immutable once set. Existing Releases with an explicit uniqueName are unaffected. When left empty, the Target controller uses the parent Component name as the effective deduplication key (visible in status.effectiveUniqueName); the field itself remains empty in storage.

**New status field — Release.Status.EffectiveUniqueName:** Read-only, set by the Release controller after CV resolution. Shows the deduplication key actually in use.

**New validation — Release.Spec.ComponentVersionRef.Name:** Now explicitly rejected as empty at create and update time, giving an immediate API error instead of a silent controller failure.

**Conversion:** `EffectiveUniqueName` is added to the internal `ReleaseStatus` type and to the generated conversion functions in `zz_generated.conversion.go` (minimal hand-edit; full codegen not needed for a single string field).

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Releases may omit UniqueName; an EffectiveUniqueName is derived and persisted for deduplication.
  * Targets now perform deterministic, priority-aware release resolution with bidirectional anti‑affinity filtering and expose a resolved condition with human-readable messages.
* **Bug Fixes**
  * Prevents creating per-release render tasks when all bindings are filtered or none apply.
* **Tests**
  * Added/updated tests for deduplication, priority/tiebreaking, anti‑affinity, resolution outcomes, and validation behavior.

### DIFF
--- a/api/solar/release_rest_test.go
+++ b/api/solar/release_rest_test.go
@@ -17,16 +17,16 @@ import (
 
 var _ = Describe("Release REST", func() {
 	Describe("Validate (create path)", func() {
-		It("rejects an empty UniqueName", func() {
+		It("rejects an empty componentVersionRef.name", func() {
 			r := &solar.Release{
 				Spec: solar.ReleaseSpec{
-					ComponentVersionRef: corev1.LocalObjectReference{Name: "kyverno-v1"},
-					UniqueName:          "",
+					ComponentVersionRef: corev1.LocalObjectReference{Name: ""},
+					UniqueName:          "kyverno",
 				},
 			}
 			errs := r.Validate(context.Background())
 			Expect(errs).NotTo(BeEmpty())
-			Expect(errs[0].Field).To(Equal("spec.uniqueName"))
+			Expect(errs[0].Field).To(Equal("spec.componentVersionRef.name"))
 		})
 
 		It("accepts a non-empty UniqueName", func() {
--- a/pkg/controller/target_controller.go
+++ b/pkg/controller/target_controller.go
@@ -19,6 +19,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/events"
@@ -35,6 +36,7 @@ const (
 	targetFinalizer = "solar.opendefense.cloud/target-finalizer"
 
 	ConditionTypeRegistryResolved = "RegistryResolved"
+	ConditionTypeReleasesResolved = "ReleasesResolved"
 	ConditionTypeReleasesRendered = "ReleasesRendered"
 	ConditionTypeBootstrapReady   = "BootstrapReady"
 )
@@ -42,11 +44,14 @@ const (
 var ErrReleaseNotRenderedYet = errors.New("release is not rendered yet")
 
 type releaseInfo struct {
-	name     string
-	release  *solarv1alpha1.Release
-	cv       *solarv1alpha1.ComponentVersion
-	rtName   string
-	chartURL string
+	// bindingKey is "<namespace>/<name>" of the originating ReleaseBinding, used as a
+	// deterministic tiebreaker when two releases share the same priority.
+	bindingKey string
+	name       string
+	release    *solarv1alpha1.Release
+	cv         *solarv1alpha1.ComponentVersion
+	rtName     string
+	chartURL   string
 }
 
 type TargetReconciler struct {
@@ -202,7 +207,12 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	if len(bindingList.Items) == 0 {
 		log.V(1).Info("No ReleaseBindings found for target")
-		if condErr := r.setCondition(ctx, target, ConditionTypeReleasesRendered, metav1.ConditionFalse, "NoBindings",
+		if condErr := r.setCondition(ctx, target, ConditionTypeReleasesRendered, metav1.ConditionFalse, "NoReleaseBindings",
+			"No ReleaseBindings found for this target"); condErr != nil {
+			return ctrl.Result{}, condErr
+		}
+
+		if condErr := r.setCondition(ctx, target, ConditionTypeReleasesResolved, metav1.ConditionFalse, "NoReleaseBindings",
 			"No ReleaseBindings found for this target"); condErr != nil {
 			return ctrl.Result{}, condErr
 		}
@@ -272,11 +282,28 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 		rtName := releaseRenderTaskName(rel.Name, target.Name, rel.GetGeneration())
 		releases = append(releases, releaseInfo{
-			name:    rel.Name,
-			release: rel,
-			cv:      cv,
-			rtName:  rtName,
+			bindingKey: binding.Namespace + "/" + binding.Name,
+			name:       rel.Name,
+			release:    rel,
+			cv:         cv,
+			rtName:     rtName,
 		})
+	}
+
+	// Resolve conflicts: deduplicate by uniqueName (priority wins) and apply anti-affinity rules.
+	var skipped []string
+	releases, skipped = resolveReleaseConflicts(releases)
+	if condErr := r.setResolvedCondition(ctx, target, skipped); condErr != nil {
+		return ctrl.Result{}, condErr
+	}
+
+	if len(releases) == 0 && !pendingDeps {
+		if condErr := r.setCondition(ctx, target, ConditionTypeReleasesRendered, metav1.ConditionFalse, "AllReleaseBindingsFiltered",
+			"All ReleaseBindings were filtered out by the release resolver (uniqueName conflicts or anti-affinity rules)"); condErr != nil {
+			return ctrl.Result{}, condErr
+		}
+
+		return ctrl.Result{}, nil
 	}
 
 	// Create per-release RenderTasks (one per target+release pair).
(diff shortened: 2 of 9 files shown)


dermorz:
@coderabbitai

Regarding your last nitpick comment:
`False/NoBindings` is the right value here. If we used `True/NoConflicts`, operators couldn't tell the difference between "there were no bindings at all" and "there were bindings but none conflicted" — both would look the same. The `False` status deliberately matches `ReleasesRendered=False/NoBindings` set just above it, so both conditions point to the same root cause.

pkg/controller/target_controller.go
```diff
@@ -272,13 +282,30 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
+	// Resolve conflicts: deduplicate by uniqueName (priority wins) and apply anti-affinity rules.
+	var skipped []string
+	releases, skipped = resolveReleaseConflicts(releases)
+	if condErr := r.setResolvedCondition(ctx, target, skipped); condErr != nil {
+		return ctrl.Result{}, condErr
+	}
+
+	if len(releases) == 0 && !pendingDeps {
+		if condErr := r.setCondition(ctx, target, ConditionTypeReleasesRendered, metav1.ConditionFalse, "AllBindingsFiltered",
```
yocaba: `AllBindingsFiltered` might be misleading since there are also `RegistryBindings`. Better change to `AllReleaseBindingsFiltered`?

pkg/controller/target_controller.go
```diff
@@ -272,13 +282,30 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			name:       rel.Name,
+			release:    rel,
+			cv:         cv,
+			rtName:     rtName,
 		})
 	}
 
+	// Resolve conflicts: deduplicate by uniqueName (priority wins) and apply anti-affinity rules.
+	var skipped []string
+	releases, skipped = resolveReleaseConflicts(releases)
```
yocaba, line 295: I was confused by `skipped` first, would be `filtered` a better match also according to the Condition Status text below? Either way, we should be consistent and go with "skipped" or "filtered" also in the other user-facing texts.

pkg/controller/target_controller.go
```diff
@@ -465,6 +492,123 @@ func (r *TargetReconciler) setCondition(ctx context.Context, target *solarv1alph
...
+		return releases, nil
+	}
+
+	// Step A: uniqueName deduplication.
+	// Releases with an empty uniqueName are never deduplicated.
+	namedGroups := map[string][]releaseInfo{}
+	var accepted []releaseInfo
+
+	for _, ri := range releases {
+		if ri.release.Spec.UniqueName == "" {
```
yocaba: That can lead to conflicts, can't it? Is that even a valid use case that the `UniqueName` is empty or if it is empty we need a fallback and derive it somehow?

pkg/controller/target_controller_test.go
```diff
@@ -290,6 +291,172 @@ var _ = Describe("TargetController", Ordered, func() {
...
+
+			// Target should reflect the conflict resolution
+			Eventually(func() bool {
+				t := &solarv1alpha1.Target{}
+				if err := k8sClient.Get(ctx, client.ObjectKeyFromObject(target), t); err != nil {
+					return false
+				}
+				cond := apimeta.FindStatusCondition(t.Status.Conditions, ConditionTypeReleasesResolved)
+
+				return cond != nil && cond.Status == metav1.ConditionTrue && cond.Reason == "Resolved"
```
yocaba: We could check the message as well, at least that the filtered release is mentioned.

pkg/controller/target_controller_test.go
```diff
@@ -290,6 +291,172 @@ var _ = Describe("TargetController", Ordered, func() {
...
+			Eventually(func() error {
+				return k8sClient.Get(ctx, client.ObjectKey{Name: alphaRTName, Namespace: ns.Name}, &solarv1alpha1.RenderTask{})
+			}, eventuallyTimeout).Should(Succeed())
+
+			zetaRTName := releaseRenderTaskName("rel-tiebreak-z", "test-resolver-tiebreak", 1)
+			Consistently(func() bool {
+				err := k8sClient.Get(ctx, client.ObjectKey{Name: zetaRTName, Namespace: ns.Name}, &solarv1alpha1.RenderTask{})
+				return apierrors.IsNotFound(err)
+			}, 2*time.Second, 100*time.Millisecond).Should(BeTrue())
+		})
```
yocaba, line 392: Is the status condition check skipped on purpose?

yocaba (commented):
Overall looks good to me, just a few naming comments and a (recurring :)) clarification question regarding the semantics around `UniqueName`.

dermorz:
@coderabbitai review

pkg/controller/target_controller.go
dermorz: Hm, i used "skipped" in the code because I felt it reads more naturally, but "filtered" in user facing messages.

If you persist, I can change it to "filtered" everywhere.

yocaba: User-facing texts are more important to me and since they are consistent now, I'm fine with how it is now.

api/solar/release_rest_test.go
```diff
@@ -17,16 +17,26 @@ import (
...
+					UniqueName:          "kyverno",
 				},
 			}
 			errs := r.Validate(context.Background())
 			Expect(errs).NotTo(BeEmpty())
-			Expect(errs[0].Field).To(Equal("spec.uniqueName"))
+			Expect(errs[0].Field).To(Equal("spec.componentVersionRef.name"))
+		})
+
+		It("accepts an empty UniqueName", func() {
```
yocaba: Why would we test that an empty string is fine? If we have the field declared, why should an empty string not work?

pkg/controller/target_controller.go
```diff
@@ -207,6 +212,11 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, condErr
 		}
 
+		if condErr := r.setCondition(ctx, target, ConditionTypeReleasesResolved, metav1.ConditionFalse, "NoBindings",
```
yocaba: It should be `NoReleaseBindings`?

pkg/controller/target_controller.go
```diff
@@ -465,6 +492,125 @@ func (r *TargetReconciler) setCondition(ctx context.Context, target *solarv1alph
...
+
+	// Step A: uniqueName deduplication.
+	// When UniqueName is empty, fall back to the parent Component name from the CV.
+	namedGroups := map[string][]releaseInfo{}
+
+	for _, ri := range releases {
+		uniqueName := ri.release.Spec.UniqueName
+		if uniqueName == "" {
+			uniqueName = ri.cv.Spec.ComponentRef.Name
+		}
```
yocaba, line 520: We should use `Status.EffectiveUniqueName` here, shouldn't we?

pkg/controller/target_controller_test.go
```diff
@@ -70,7 +72,7 @@ var _ = Describe("TargetController", Ordered, func() {
 				},
 				Spec: solarv1alpha1.ReleaseSpec{
 					ComponentVersionRef: corev1.LocalObjectReference{Name: "my-cv"},
-					UniqueName:          "my-component",
+					UniqueName:          name,
```
yocaba: I guess it is better to stick we a unique name in the test that is different from the resources name to ensure the release resolvement actually considers the unique name. Or leave it empty in the default instance creation?

yocaba (changes requested):

pkg/controller/target_controller.go
dermorz: Good catch! Missed that one.

api/solar/release_rest_test.go
dermorz: That was a leftover from earlier behavior. I will drop it.

pkg/controller/target_controller.go
dermorz: Using `status.effectiveUniqueName` here would introduce a controller odering dependency. The target controller would silently fall back to an empty string if the release controller hasn't reconciled yet and set that field. That's why reading directly from `cv.Spec.ComponentRef.Name` keeps the deduplication self-contained.

We also already have fetched cv, so there is no extra query.

yocaba: Understood. May be nice to put the logic how the effective name is determined at one place and use it for setting the Status and for release resolving. But I also expect the tests to catch a regression, meaning if the logic diverges.

yocaba (approved):
lgtm

PR #497, opened 2026-05-05
chore: switching runners

AlexHardAtWork:
## What
Migrate to GitHub Runners, fix tests for different filesystem situations.

Additionally this fixes a E2E test fail.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **Chores**
  * Standardized CI/CD runners to Ubuntu 24.04 across workflows (Docker, Go, Helm, docs, release, and CI checks).
  * Adjusted the signing tool version used in publishing workflows.
  * Improved test coverage reporting to aggregate per-package results into a combined coverage profile.

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   pr-title:
     name: PR Title
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: amannn/action-semantic-pull-request@v6
         env:
@@ -31,7 +31,7 @@ jobs:
 
   commitlint:
     name: Commit Messages
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
         with:
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -88,7 +88,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Install cosign
-        uses: sigstore/cosign-installer@v4.1.1
+        uses: sigstore/cosign-installer@v4.1.2
       - name: Sign image with cosign
         env:
           COSIGN_EXPERIMENTAL: 1
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -42,7 +42,7 @@ jobs:
       github.event_name == 'push' ||
       (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ok-to-image')) ||
       (github.event_name == 'release' && github.event.action == 'published')
-    runs-on: arc-scale-set
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
       - uses: docker/metadata-action@v6
@@ -88,7 +88,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Install cosign
-        uses: sigstore/cosign-installer@v4.1.2
+        uses: sigstore/cosign-installer@v4.1.1
       - name: Sign image with cosign
         env:
           COSIGN_EXPERIMENTAL: 1
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -42,7 +42,7 @@ jobs:
       github.event_name == 'push' ||
       (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ok-to-image')) ||
       (github.event_name == 'release' && github.event.action == 'published')
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
       - uses: docker/metadata-action@v6
--- a/Makefile
+++ b/Makefile
@@ -48,10 +48,14 @@ lint-no-golangci: $(ADDLICENSE) shellcheck  ## Run linters but not golangci-lint
 
 .PHONY: test
 test: $(SETUP_ENVTEST) $(GINKGO) ocm-transfer-demo ## Run all tests
+	mkdir -p $(BUILD_PATH)/coverdata
 	OCM=$(OCM) \
+	GOCOVERDIR=$(BUILD_PATH)/coverdata \
 	KUBEBUILDER_ASSETS="$(shell $(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" \
-	$(GINKGO) -r -cover --fail-fast --require-suite -covermode count --output-dir=$(BUILD_PATH) -coverprofile=solar.full.coverprofile $(testargs)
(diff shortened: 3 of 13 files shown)


AlexHardAtWork:
@coderabbitai:
> 🧹 Nitpick comments (2)

The pipeline has been addressed, the Makefile common.mk will be handled $elsewhere.

pkg/discovery/pipeline/pipeline.go
```diff
@@ -71,25 +70,41 @@ func NewPipeline(namespace string, registries *discovery.RegistryProvider, webho
...
+		}
+	}
+
+	p.qualifier = qualifier.NewQualifier(registries, namespace, repoEvents, filterInput, errChan, discovery.WithLogger[discovery.RepositoryEvent, discovery.ComponentVersionEvent](log))
+
+	p.filter = handler.NewFilter(solarClient, namespace, filterInput, handlerInput, errChan, discovery.WithLogger[discovery.ComponentVersionEvent, discovery.ComponentVersionEvent](log))
+
+	p.handler = handler.NewHandler(registries, handlerInput, writerInput, errChan, discovery.WithLogger[discovery.ComponentVersionEvent, discovery.WriteAPIResourceEvent](log), discovery.WithRateLimiter[discovery.ComponentVersionEvent, discovery.WriteAPIResourceEvent](time.Second, 1))
+
+	p.writer = apiwriter.NewAPIWriter(solarClient, namespace, registries, writerInput, errChan, discovery.WithLogger[discovery.WriteAPIResourceEvent, any](log))
```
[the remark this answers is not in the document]
AlexHardAtWork: NewFilter and NewAPIWriter only store the nil client in a struct field — their constructors don't
  dereference it. The nil is only reachable through Filter.Process() / APIWriter.Process(), which are the
   default processors on the Runner. The second option pass (line 108-110) replaces those processors with
   fakes via WithFilterProcessor/WithWriterProcessor, so the real Process() methods are never called and
  the nil client is never dereferenced.

  WithoutSolarClient() is a test-only escape hatch that's always paired with processor overrides. Adding
  conditional construction of filter/writer plus Start()-time validation would add significant complexity
   (nil checks in Start(), Stop(), option guards) to protect against a misuse scenario that doesn't
  exist. The nil guards on the With* options already make the API safe.

Makefile
```diff
@@ -47,10 +47,14 @@ lint-no-golangci: $(ADDLICENSE) shellcheck  ## Run linters but not golangci-lint
...
+	mkdir -p $(BUILD_PATH)/coverdata
 	OCM=$(OCM) \
+	GOCOVERDIR=$(BUILD_PATH)/coverdata \
 	KUBEBUILDER_ASSETS="$(shell $(SETUP_ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" \
-	$(GINKGO) -r -cover --fail-fast --require-suite -covermode count --output-dir=$(BUILD_PATH) -coverprofile=solar.full.coverprofile $(testargs)
-	@grep -v /solar/api solar.full.coverprofile > solar.coverprofile
+	$(GINKGO) -r -cover --fail-fast --require-suite -covermode count --output-dir=$(BUILD_PATH) -coverprofile=solar.full.coverprofile --keep-separate-coverprofiles $(testargs)
+	@echo 'mode: count' > $(BUILD_PATH)/solar.full.coverprofile && \
+	 cat $(BUILD_PATH)/*_solar.full.coverprofile 2>/dev/null | grep -v '^mode:' >> $(BUILD_PATH)/solar.full.coverprofile || true
+	@grep -v /solar/api $(BUILD_PATH)/solar.full.coverprofile > solar.coverprofile
```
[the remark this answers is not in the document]
AlexHardAtWork: finding addressed.

AlexHardAtWork:
> 🧹 Nitpick comments (1)
> > Makefile (1)> `50-57`: _⚡ Quick win_
> > > **`GOCOVERDIR` data is collected but never merged into the coverage report.**
> > > ...

 @coderabbitai: The tests invoke the external ocm binary as a subprocess, but it's a third-party
   tool — not built with -cover from this repo's source. No instrumented subprocess writes coverage data
  to GOCOVERDIR. For the test binaries themselves, ginkgo's -coverprofile already captures coverage in
  text format, so any binary covdata in GOCOVERDIR would be redundant. Adding a merge step would either
  be a no-op (empty directory) or cause double-counting. If anything, the GOCOVERDIR/mkdir -p coverdata
  lines are dead code that could be removed, but that's not what the finding asks for.

pkg/discovery/pipeline/pipeline.go
```diff
@@ -150,32 +165,52 @@ func (p *Pipeline) Stop(ctx context.Context) error {
 	return err
 }
 
+// WithoutSolarClient skips creation of the Kubernetes Solar API client.
+// Callers must also provide WithFilterProcessor and WithWriterProcessor
+// overrides, because the default Filter and APIWriter processors will
+// panic on a nil client when they receive the first event.
+func WithoutSolarClient() Option {
```
trevex: Does it make sense to "invert" this behavior.

Instead of auto-constructing, we instead take a `WithSolarClient` and properly validate that either SolarClient is provided or With*Processor are properly provided.

AlexHardAtWork: Moved out to #507

.github/workflows/docker.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
       github.event_name == 'push' ||
       (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ok-to-image')) ||
       (github.event_name == 'release' && github.event.action == 'published')
-    runs-on: arc-scale-set
+    runs-on: ubuntu-latest
```
yocaba: We better pin the version to have reproducible builds or at least avoid that they start breaking out of a sudden when the latest tag moved.

The latest stable version is 24.04. 26.04 is still in the making: https://github.com/actions/runner-images/issues/13964

AlexHardAtWork: @yocaba : fixed to ubuntu-24.04

One note: Renovate can't pick up these labels, we need to bump manually when ubuntu-26.04 is out.

.github/workflows/conventional-commits.yml
```diff
@@ -11,7 +11,7 @@ permissions:
 jobs:
   pr-title:
     name: PR Title
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
yocaba, line 14: Just for clarification: My concern in my previous comment was regarding GitHub actions related to our builds. I don't have concerns for common GitHub actions like conventional commit checks, issue adding or release drafter. But since we have the pinned versions in this repos for build-related GitHub actions, it doesn't hurt to have them for the common GitHub actions as well though.

pkg/discovery/pipeline/pipeline.go
AlexHardAtWork: #507 is merged.

yocaba (approved):

.github/workflows/conventional-commits.yml
AlexHardAtWork: Fixed Renovate.

PR #499, opened 2026-05-05
feat: polish of ui poc: added impersonation, updated docs, dev-setup,…

### LINKED ISSUE #432: Polish UI prototype
Follow up to take care of loose ends of #166 

AC:
* Finish drop-down role selection implementation
* Update docs, e.g. developer guidance on frontend bits, updates to ADR as required
* Polish code, ci, ...

rebEllieous:
… helm-template and ui-e2e test

## What
Implements user impersonation ("Preview as") in the solar-ui BFF and frontend.
Closes #432 

## Why
Platform admins need to preview the UI as other users to validate that RBAC restrictions are working correctly, without logging out and back in as a different user.

## Testing
- Manual: logged in as admin@solar.local against `make ui-dev` cluster, verified /api/auth/impersonation-targets returns the configured personas, impersonation activates/clears correctly, permissions reflect the persona's RBAC, and isAdmin remains true on the real identity while impersonating
- Playwright e2e: new impersonation.spec.ts covering unauthenticated 401s, admin access, the activate → check permissions → clear lifecycle, and invalid input rejection. run via `make ui-test-e2e`

## Notes for reviewers
**How admin access works:**
The BFF uses its own service-account credentials to list ClusterRoleBindings labeled solar.opendefense.cloud/admin=true and checks the session's username/groups against subjects. This avoids issuing a SelfSubjectAccessReview as the logged-in user (which would fail in impersonate auth-mode) and is immune to active impersonation state.
**New RBAC resources (via Helm):**
- admin-clusterrolebinding.yaml: solar-ui:admin ClusterRole + CRB, populated from values.ui.admin.subjects with admin label used to distinguish admins in the BFF
- impersonation-clusterrolebinding.yaml: solar-ui:impersonate:$username ClusterRoles + CRBs, populated from values.ui.impersonation.targets with impersonatable label
- rbac.yaml: grants the BFF SA list on clusterroles and clusterrolebindings (needed for admin check and impersonatable discovery)

## Checklist
- [*] Tests added/updated
- [*] No breaking changes (or upgrade path documented above)
- [*] Readable commit history (squashed and cleaned up as desired)
- [*] AI code review considered and comments resolved

## Summary by CodeRabbit

* **New Features**
  * Administrators can now impersonate other users to preview and test their experience
  * Admin role support integrated with Kubernetes RBAC
  * Routes and navigation now protected based on user permissions

* **Documentation**
  * Added UI access control configuration guide

* **Tests**
  * Added impersonation workflow tests

### DIFF
--- a/pkg/ui/api/handler.go
+++ b/pkg/ui/api/handler.go
@@ -39,6 +39,7 @@ var resourceMap = map[string]schema.GroupVersionResource{
 // Handler serves the K8s API proxy routes.
 type Handler struct {
 	baseConfig   *rest.Config
+	clientset    kubernetes.Interface
 	sessionStore *session.Store
 	authProvider auth.Provider
 	log          logr.Logger
@@ -58,8 +59,14 @@ func NewHandler(kubeconfig string, store *session.Store, provider auth.Provider,
 		return nil, fmt.Errorf("failed to create kubernetes config: %w", err)
 	}
 
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kubernetes clientset: %w", err)
+	}
+
 	return &Handler{
 		baseConfig:   cfg,
+		clientset:    clientset,
 		sessionStore: store,
 		authProvider: provider,
 		log:          log.WithName("api"),
@@ -79,15 +86,12 @@ func (h *Handler) clientFor(r *http.Request) (dynamic.Interface, error) {
 // Uses the BFF's own service-account credentials so the check is immune to
 // any active session-level impersonation override.
 func (h *Handler) isAdminUser(ctx context.Context, sess *session.Data) bool {
-	clientset, err := kubernetes.NewForConfig(h.baseConfig)
-	if err != nil {
-		return false
-	}
-
-	bindingList, err := clientset.RbacV1().ClusterRoleBindings().List(ctx, metav1.ListOptions{
+	bindingList, err := h.clientset.RbacV1().ClusterRoleBindings().List(ctx, metav1.ListOptions{
 		LabelSelector: adminLabel + "=true",
 	})
 	if err != nil {
+		h.log.Error(err, "failed to list admin ClusterRoleBindings")
+
 		return false
 	}
 
@@ -373,12 +377,7 @@ type ImpersonationTarget struct {
 // resourceNames in those rules define the username and group membership of the
 // persona respectively.
 func (h *Handler) listImpersonationTargets(ctx context.Context) ([]ImpersonationTarget, error) {
-	clientset, err := kubernetes.NewForConfig(h.baseConfig)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create kubernetes clientset: %w", err)
-	}
-
-	roleList, err := clientset.RbacV1().ClusterRoles().List(ctx, metav1.ListOptions{
+	roleList, err := h.clientset.RbacV1().ClusterRoles().List(ctx, metav1.ListOptions{
 		LabelSelector: impersonatableLabel + "=true",
 	})
 	if err != nil {
@@ -476,7 +475,12 @@ func (h *Handler) HandleImpersonate() http.HandlerFunc {
 			return
 		}
 
-		h.sessionStore.SetImpersonation(r, target.Username, target.Groups)
+		if !h.sessionStore.SetImpersonation(r, target.Username, target.Groups) {
+			http.Error(w, "no session found", http.StatusUnauthorized)
+
+			return
+		}
+
 		w.WriteHeader(http.StatusNoContent)
 	}
 }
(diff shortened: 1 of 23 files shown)


pkg/ui/api/handler.go
```diff
@@ -69,10 +74,82 @@ func (h *Handler) clientFor(r *http.Request) (dynamic.Interface, error) {
 	return dynamic.NewForConfig(cfg)
 }
 
-// HandleMe returns the current user info.
+// isAdminUser returns true when the session user is a subject of any
+// ClusterRoleBinding labeled solar.opendefense.cloud/admin=true.
+// Uses the BFF's own service-account credentials so the check is immune to
+// any active session-level impersonation override.
+func (h *Handler) isAdminUser(ctx context.Context, sess *session.Data) bool {
+	clientset, err := kubernetes.NewForConfig(h.baseConfig)
```
dermorz: This creates a new clientset every time `isAdminUser()` is called. I think we can create a clientset on construction  time and cache it on `Handler`.

pkg/ui/api/handler.go
```diff
@@ -69,10 +74,82 @@ func (h *Handler) clientFor(r *http.Request) (dynamic.Interface, error) {
...
+	if err != nil {
+		return false
```
dermorz: Since we do not pass the `err` that is not `nil` anywhere, we should probably log it here befor returning.

(Or if clientset creation is moved from this func to Handler construction, log it there)

pkg/ui/api/handler.go
```diff
@@ -69,10 +74,82 @@ func (h *Handler) clientFor(r *http.Request) (dynamic.Interface, error) {
...
+	}
+
+	bindingList, err := clientset.RbacV1().ClusterRoleBindings().List(ctx, metav1.ListOptions{
+		LabelSelector: adminLabel + "=true",
+	})
+	if err != nil {
+		return false
```
dermorz, line 95: Same as above: The non-`nil` error is not passed anywhere so we should probably log it before return.

pkg/ui/api/handler.go
```diff
@@ -211,3 +288,203 @@ func (h *Handler) HandleSSE() http.HandlerFunc {
...
+}
+
+// listImpersonationTargets reads ClusterRoles labeled
+// solar.opendefense.cloud/impersonatable=true using the BFF's own service
+// account credentials (not the logged-in user's). Each ClusterRole is expected
+// to grant the impersonate verb on both the "users" and "groups" resources; the
+// resourceNames in those rules define the username and group membership of the
+// persona respectively.
+func (h *Handler) listImpersonationTargets(ctx context.Context) ([]ImpersonationTarget, error) {
+	clientset, err := kubernetes.NewForConfig(h.baseConfig)
```
dermorz: Same as on `isAdminUser()`. If we have a clientset cached on `Handler`, we could just use that one.

pkg/ui/server.go
```diff
@@ -92,6 +92,17 @@ func NewServer(cfg Config, log logr.Logger) (*Server, error) {
 	// SSE events
 	mux.Handle("GET /api/namespaces/{namespace}/events", requireAuth(k8sHandler.HandleSSE()))
 
+	// Permissions — uses SelfSubjectRulesReview with caller's own credentials
+	mux.Handle("GET /api/namespaces/{namespace}/permissions", requireAuth(k8sHandler.HandlePermissions()))
+
+	// Impersonation
+	// Admin check is SAR-based (impersonate verb on users), matching isAdmin in /auth/me.
```
dermorz: I think this comment is a leftover from the PoC, right? It's not SAR-based anymore but based on the admin label on ClusterRolebindings.

pkg/ui/api/handler.go
```diff
@@ -211,3 +288,203 @@ func (h *Handler) HandleSSE() http.HandlerFunc {
...
+			}
+		}
+
+		if target == nil {
+			http.Error(w, "unknown impersonation target", http.StatusBadRequest)
+
+			return
+		}
+
+		h.sessionStore.SetImpersonation(r, target.Username, target.Groups)
```
dermorz: The boolean return value of `SetImpersonation()` is not evaluated. Following the http status code 204 is always set, when it worked and if it didn't work.

Maybe we want to respond with 401 if SetImposonation fails (no session found).

pkg/ui/api/handler.go
```diff
@@ -211,3 +288,203 @@ func (h *Handler) HandleSSE() http.HandlerFunc {
...
+		w.WriteHeader(http.StatusNoContent)
+	}
+}
+
+// HandleClearImpersonation removes the impersonation override from the session.
+func (h *Handler) HandleClearImpersonation() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		h.sessionStore.ClearImpersonation(r)
```
dermorz: Same as above with `SetImpersonation()`. Let's not ignore the return value here and set the http status code accordingly.

pkg/ui/api/handler.go
```diff
@@ -69,10 +74,82 @@ func (h *Handler) clientFor(r *http.Request) (dynamic.Interface, error) {
...
+
+		next(w, r)
+	}
+}
+
+// HandleMe returns the current user info, including an isAdmin flag derived
+// from membership in a ClusterRoleBinding labeled solar.opendefense.cloud/admin=true
 func (h *Handler) HandleMe() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		h.sessionStore.GetJSON(w, r)
```
dermorz, line 75: This was the only call to `GetJSON()`. Let's remove it from `store.go`. 🚮

dermorz (changes requested):
I focused on the Go portion of this PR and trusted the tests with the Frontend part.

dermorz (approved):
lgtm 👍

PR #515, opened 2026-05-08
chore: consolodate registry types

### LINKED ISSUE #437: Refactor internal use of registry
We should try to use one internal Registry Go type across the codebase.

PoC started in #436

AC:
* revisit all Registry types
* consolidate as much as possible and reasonable

rebEllieous:
## What
Consolidate the internal `discovery.Registry` struct into the `solar.Registry` API type.
Closes #437

## Why
The discovery worker previously maintained its own parallel `Registry` struct loaded from a static Helm ConfigMap. This duplicated information already available or belongs in the `solar.Registry` API object, and required operators to keep two sources of truth in sync. By moving those fields onto `RegistrySpec` and deriving everything from the API object at runtime, the discovery worker becomes purely Kubernetes-native, with no separate config file.

## Testing
- Unit tests updated
- E2E fixtures updated to use `Registry` objects instead of Helm values for scan/webhook configuration

## Notes for reviewers
- **API change**: `RegistrySpec` gains `flavor`, `webhookPath`, and `scanInterval` fields. These are additive and optional; existing `Registry` objects without them continue to work.
- The `solar-discovery` Helm chart no longer renders a ConfigMap for registry config or mounts it into the deployment. The `values.yaml` registry list entries have been removed.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Registry CRDs now support flavor, webhookPath, and scanInterval; discovery loads registries from the Kubernetes API and CLI no longer needs a local config file.
  * Server-side validation added for Registry webhook/scan fields.

* **Bug Fixes**
  * RBAC updated to allow discovery to read Registry resources and referenced Secrets.

* **Documentation**
  * API and user docs updated for new registry fields and release deduplication.

* **Tests / Chores**
  * Tests and e2e fixtures updated for the new Registry model; chart defaults/configmap for registries removed.


test/fixtures/e2e/zot-discovery-registry-webhook.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: solar.opendefense.cloud/v1alpha1
+kind: Registry
+metadata:
+  name: zot-webhook
```
dermorz, line 4: In `test/e2e/e2e_test.go` L99 and L299 the following command is built to delete this Registry:
`kubectl delete registry -n testns --selector app=zot-discovery-webhook --ignore-not-found`

But on this Registry Resource there are no labels defined. So `kubectl delete registry` does not match any resource and is failing silently because of `--ignore-not-found`.

Either add the app-label here or change the `kubectl delete registry` to "by name".

api/solar/v1alpha1/registry_types.go
```diff
@@ -24,6 +26,19 @@ type RegistrySpec struct {
...
+	// Required when WebhookPath is set.
+	// +optional
+	Flavor string `json:"flavor,omitempty"`
+	// WebhookPath is the HTTP path on which the discovery worker listens for
+	// push notifications from this registry. Leave empty to disable webhook-based
+	// discovery and rely on polling via ScanInterval instead.
+	// +optional
+	WebhookPath string `json:"webhookPath,omitempty"`
+	// ScanInterval controls how often the discovery worker performs a full scan
+	// of this registry. Defaults to 24h when unset.
```
dermorz: `ScanInterval` now defaults to empty, so we need to change the docs. Here on this field and everywhere else, so the users know they have to set a `ScanInterval` to enable scanning at all.

pkg/discovery/registry_provider.go
```diff
@@ -4,140 +4,122 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to read secret %q for registry %q: %w", reg.Spec.SolarSecretRef.Name, reg.Name, err)
+		}
 
-	var ymlConfig RegistryProviderConfig
-	for _, reg := range p.registry {
-		ymlConfig.Registries = append(ymlConfig.Registries, reg)
+		creds[reg.Name] = &RegistryCredentials{
+			Username: string(secret.Data[SecretKeyUsername]),
+			Password: string(secret.Data[SecretKeyPassword]),
```
dermorz: If `secret.Data` does not contain the keys `"username"` or `"password"`, they default to emptystring. This would result in silently setting wrong/missing `RegistryCredentials` and later failures when communicating with Registries.

I think we should add either explicit key existence checks or a value length check here for early failure on missing credentials.

dermorz (changes requested):
Apart from those 3 comments I noticed that `Registry` resources are only read on discovery startup (`init()`). We probaly should open a new issue for watching `Registry` resources during runtime, because right now adding or changing `Registries` or their credentials would only be read on discovery restart. (But I think this is not subject of this issue/PR)

dermorz (approved):
lgtm 🚀

PR #517, opened 2026-05-08
fix(docs): api ref list types

dermorz:
## What

Closes #366

Fix the API reference generator to emit documentation sections for `*List` resource types, resolving broken anchor links (e.g. `[TargetList](#targetlist)`) throughout `docs/user-guide/api-reference.md`. Add a lint check to detect upstream template drift on future crd-ref-docs version bumps.

## Why

`crd-ref-docs` only renders a type when it has a `// +kubebuilder:object:root=true` marker or is referenced by a field in another type. List types satisfy neither condition — nothing declares a field of type `TargetList`, and this project uses k8s code-gen markers rather than kubebuilder ones. The generator still emitted cross-reference links such as _Appears in: [TargetList](#targetlist)_ under the singular types, but the corresponding anchor never existed, producing broken links flagged by markdownlint as `MD051`.

The fix provides custom templates via `--templates-dir`. The only change from the crd-ref-docs defaults is one line in `type.tpl`: the render guard is extended with `(hasSuffix "List" $type.Name)` using sprig (already in the template function map).

The drift check (`hack/check-crd-ref-docs-templates.sh`) stores SHA256 checksums of the four upstream template files at the pinned version and fails `lint-no-golangci` if they change, so a crd-ref-docs upgrade never silently invalidates the patch.

## Testing

- `make docs-crd-ref` — all ten `*List` anchor links now resolve to their sections.
- `bash hack/check-crd-ref-docs-templates.sh` — exits 0 against the pinned version.
- Manually corrupted a checksum — script exits 1 with a diff and remediation hint.
- `make lint-no-golangci` — passes end-to-end.

## Notes for reviewers

No CRD/API changes, no RBAC changes. When upgrading `crd-ref-docs` in `tools.lock`, `make lint-no-golangci` will fail with a diff if the upstream templates changed — follow the printed instructions to re-sync and update `.upstream-checksums`.

## Checklist

- [x] ~~Tests added/updated~~ n/a

## Summary by CodeRabbit

* **Documentation**
  * Expanded API reference with missing List resource sections and clarified Release fields (uniqueName and new effectiveUniqueName).
  * Fixed minor API doc comment inconsistencies.

* **Chores**
  * Docs generation now uses an additional templates directory.
  * Added a check that verifies local documentation templates against upstream checksums and fails CI on mismatches.

* **Style**
  * Added new markdown templates to standardize CRD reference output.


yocaba (approved):
lgtm

Is that an opportunity to fix or at least address that gap with List types upstream in crd-ref-docs?

dermorz:
> Is that an opportunity to fix or at least address that gap with List types upstream in crd-ref-docs?

I will check if there is an upstream issue or if I can address it.

dermorz:
@yocaba https://github.com/elastic/crd-ref-docs/issues/60 already exists and there is already someone working on a non-breaking solution.

PR #553, opened 2026-05-27
feat: improved target secret rendering

### LINKED ISSUE #440: Set imagePullSecrets correctly during rendering
Currently the imagePullSecrets are not automatically set based on the information we have about registries.
However we want to make it as easily as possible for the user to get imagePullSecrets right.

There are three parts to this:
1. HelmReleases for rendered artifacts need imagePullSecrets for the "render registry"
2. The correct imagePullSecret has to be used to deploy the ComponentVersion contained in the rendered `Release`
3. The images deployed by the release should use the correct imagePullSecrets

The third use-case is probably the most complex one. The "helm values template" renders the OCI image paths, however is unaware in regards to pull secrets. We could either introduce this feature to our release values rendering, or alternatively address it directly in `ocm-kit`.

AC:
* The controller manager should have a strict mode and relaxed mode:
  * strict mode should warn/error if no RegistryBinding exists for a particular 
  * relaxed mode should assume the registry needs no authentication if no associated RegistryBinding exists
* HelmReleases contained in rendered artifacts have the correct imagePullSecrets or a mechanism, e.g. userdata, correctly sets the image pull secrets
* images deployed by a release (contained in OCM) use the correct imagePullSecrets, if the user correctly "mapped" them
* extend tests as necessary and make sure E2E tests pass
* document imagePullSecret behavior for users and operators

AlexHardAtWork:
## What

Improves the rendering for imagePullSecrets.

Closes #440 

## Testing
Unit Tests and E2E.

## Checklist
- [X] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Rendering now resolves per-resource pull-secret names via RegistryBinding and includes them in rendered manifests and release/bootstrap inputs.

* **Configuration Changes**
  * Added strict mode flag (--registry-binding-strict) to enforce RegistryBinding matches.
  * Helm values and templates updated to conditionally render pull-secret refs; RBAC updated to allow RegistryBinding access.

* **Documentation**
  * Registry credential ADR accepted; API and controller docs updated for resolved pull-secret behavior.

* **Tests**
  * Unit, renderer, and e2e tests updated for pull-secret resolution, strict mode, and spec-drift.


pkg/controller/helpers.go
```diff
@@ -89,13 +93,39 @@ func targetRenderTaskName(targetName string, bootstrapVersion int64) string {
...
+// content differs only in secretRef (due to RegistryBinding changes) get
+// unique OCI tags, preventing the renderer's exists-check from skipping a
+// necessary re-push.
+func pullSecretsTag(resolved map[string]solarv1alpha1.ResolvedResourceAccess) string {
+	keys := make([]string, 0, len(resolved))
+	for k := range resolved {
+		keys = append(keys, k)
+	}
+
+	sort.Strings(keys)
```
trevex, line 147: Makes hash deterministic 👍

pkg/controller/target_controller.go
```diff
@@ -893,6 +991,64 @@ func (r *TargetReconciler) mapRegistryToTargets(ctx context.Context, obj client.
...
+			Name:      rb.Spec.RegistryRef.Name,
+			Namespace: rb.Namespace,
+		}, reg); err != nil {
+			return nil, fmt.Errorf("failed to get Registry %s referenced by RegistryBinding %s: %w",
+				rb.Spec.RegistryRef.Name, rb.Name, err)
+		}
+
+		host := reg.Spec.Hostname
+		if existing, ok := lookup[host]; ok && existing != reg.Spec.TargetPullSecretName {
+			return nil, fmt.Errorf("conflicting RegistryBindings for host %q: pull secret %q (from %s) vs %q",
```
trevex: One bad binding currently fails the whole target render. Should we instead skip?

trevex: One bad binding currently fails the whole target render. Is this desirable?

trevex: Also: The "vs %q" side shows only a secret name, not the second binding name. The user can't locate the other offending RegistryBinding from the error

pkg/controller/target_controller.go
```diff
@@ -893,6 +991,64 @@ func (r *TargetReconciler) mapRegistryToTargets(ctx context.Context, obj client.
...
 }
 
+// buildPullSecretsLookup lists RegistryBindings for the given target, resolves
+// each bound Registry, and returns a map from registry hostname to
+// targetPullSecretName. Registries without a targetPullSecretName are included
+// with an empty string (anonymous pull).
+func (r *TargetReconciler) buildPullSecretsLookup(ctx context.Context, target *solarv1alpha1.Target) (map[string]string, error) {
+	rbList := &solarv1alpha1.RegistryBindingList{}
+	if err := r.List(ctx, rbList,
+		client.InNamespace(target.Namespace),
```
trevex, line 995: Render registry already supports cross-namespace via Target.Spec.RenderRegistryNamespace + ReferenceGrant. RegistryBinding does not. Is this something were we need a follow up ticket? Or covered elsewhere?

pkg/controller/target_controller.go
```diff
@@ -704,13 +760,50 @@ func (r *TargetReconciler) computeReleaseRenderTaskSpec(rel *solarv1alpha1.Relea
...
+			PullSecretName: pullSecret,
+		}
+	}
+
+	return resolved, nil
+}
+
+// registryHost extracts the registry host from a repository string.
+// For example, "registry.example.com:5000/foo/bar" returns "registry.example.com:5000".
+func registryHost(repository string) string {
```
trevex: `Registry.spec.hostname`: `Registry.Example.com` vs repo `registry.example.com/`. We should consider normalizing via lower case.

pkg/controller/target_controller.go
```diff
@@ -667,16 +713,26 @@ func (r *TargetReconciler) deleteOwnedRenderTasks(ctx context.Context, target *s
...
 	repo := fmt.Sprintf("%s/%s", target.Namespace, chartName)
-	tag := fmt.Sprintf("v0.0.%d", rel.GetGeneration())
 
 	var targetNamespace string
 	if rel.Spec.TargetNamespace != nil {
 		targetNamespace = *rel.Spec.TargetNamespace
 	}
 
+	resolvedResources, err := resolveResources(cv.Spec.Resources, pullSecretsByHost, r.RegistryBindingStrict)
+	if err != nil {
```
trevex, line 757: Strict-mode error has no status condition: 

Failure only surfaces in controller logs. Set a ReleasesResolved=False (or new RegistryBindingResolved)
condition on the Target with the unmatched host. Integration tests don't exercise strict mode end-to-end
either.

pkg/controller/target_controller.go
trevex: Could also be moved to `helpers.go` 🤔

pkg/controller/target_controller_test.go
```diff
@@ -205,6 +205,219 @@ var _ = Describe("TargetController", Ordered, func() {
 		})
 	})
 
+	Context("RegistryBinding pull secret resolution", Label("target"), func() {
```
trevex, line 210: Missing scenarios worth adding:
  - RegistryBinding deletion — binding removed → secret disappears from RT on next reconcile (the spec-drift
  test only covers creation).
  - RegistryBinding pointing at a non-existent Registry — currently stalls the whole target via :1014.
  Document or fix the failure surface.
  - `--registry-binding-strict=true` end-to-end — currently no integration test enables strict mode; the
  user-facing failure path is unexercised.
  - Hostname normalization — once registryHost is fixed, add an entry to the table-driven test (:719-732) for
   uppercase host and trailing-slash hostname.

trevex (dismissed):

pkg/controller/target_controller.go
AlexHardAtWork: nice spotting, fixed.

pkg/controller/target_controller.go
AlexHardAtWork: Status condition added.

pkg/controller/target_controller.go
AlexHardAtWork: Moved to #566

AlexHardAtWork: Fail/skip moved to #566.

Error message improved.

pkg/controller/target_controller.go
AlexHardAtWork: Moved to #567

pkg/controller/target_controller_test.go
AlexHardAtWork: Tests improved, strict-mode E2E moved to #568

trevex (approved):

PR #554, opened 2026-05-27
feat: implemented artifact cleanup

### LINKED ISSUE #358: Cleanup rendered resources when API resources are deleted
When SolAr API resources are deleted, their rendered artifacts (OCI images in registries, RenderTasks) should be cleaned up.

### Scenarios

1. **ReleaseBinding deleted:** The per-release RenderTask for this target+release pair should be deleted. The bootstrap RenderTask should be re-rendered without this release.
2. **Target deleted:** All RenderTasks owned by this target should be deleted. The bootstrap OCI artifact in the render registry should be cleaned up.
3. **Release deleted:** All ReleaseBindings referencing this release should be cleaned up (or blocked by deletion protection, see #308). Per-release RenderTasks referencing this release should be deleted.
4. **Registry deleted:** RegistryBindings referencing this registry should be cleaned up (or blocked by deletion protection).

### Acceptance Criteria

- [ ] Deleting a ReleaseBinding triggers bootstrap re-rendering for the target (removing the unbound release)
- [ ] Deleting a Target cleans up its RenderTasks
- [ ] OCI artifact cleanup in render registries (best-effort, as registries may be unreachable)
- [ ] Finalizers where needed to ensure cleanup before deletion completes
- [ ] Tests for each deletion scenario

### Design Notes

- The Target controller already deletes stale RenderTasks (see `cleanupStaleRenderTasks` in `target_controller.go`)
- OCI artifact deletion is registry-dependent and may not always be possible (some registries don't support tag deletion)
- Consider: should cleanup be immediate or deferred (e.g., garbage collection controller)?

### Related

- #308: Deletion protection (prevent deletion of resources that are in use)

rebEllieous:
## What
Clean up rendered OCI artifacts and `RenderTask`s when SolAr API resources are deleted.
Closes #358

## Why
Deleting a `ReleaseBinding` or `Target` previously left orphaned `RenderTask` objects and OCI images in the render registry. This implements the full deletion lifecycle.

## Testing
- Unit tests: `RenderArtifactReconciler` covered with fake client + injected `DeleteTag` stub (`pkg/controller`)
- Unit tests: `TargetReconciler` extended for `ensureRenderArtifact`, `ensureRenderBinding`, and `deleteStaleRenderBindings`
- `pkg/ociregistry.DeleteTag` tested for success and error paths
- E2E:
  - 1 artifact + 1 binding created on first render
  - Same artifact reused (2 bindings) when a second Target renders the same release
  - Artifact survives removal of one binding while another remains
  - Artifact GC'd and OCI tag deleted after last binding removed via Target deletion (finalizer path)
  - Artifact GC'd and OCI tag deleted after last ReleaseBinding deleted directly (stale-binding path)

## Notes for reviewers
**New API resources (CRD-equivalent, served via extension apiserver):**
- `RenderArtifact` records OCI coordinates of a pushed chart artifact; carries a finalizer; reconciled by `RenderArtifactReconciler`
- `RenderBinding` ref-count record linking a Target to a `RenderArtifact`; drives GC when the last binding is removed

**Helm chart:** top-level `caBundle` block added to mount a CA bundle `ConfigMap` into the controller-manager pod (`SSL_CERT_FILE`), required for TLS to self-signed registries in cluster.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added RenderArtifact and RenderBinding CRDs, typed clients/informers/listers/apply configs, and a controller that garbage-collects artifacts and performs OCI tag cleanup when no bindings remain.
  * Spec additions: pushSecretNamespace and registryFlavor; Status: chartURL and conditions.

* **Helm / Deploy**
  * Optional CA bundle mount/env and expanded RBAC for render resources.

* **Documentation**
  * API reference pages for RenderArtifact and RenderBinding added.

* **Tests**
  * New unit and E2E suites covering lifecycle, GC, OCI deletion, and auth.


pkg/controller/renderartifact_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	deleteFn := r.DeleteTag
+	if deleteFn == nil {
+		deleteFn = ociregistry.DeleteTag
+	}
+
+	auth := r.resolveAuth(ctx, artifact, registryHost)
+
+	deleteCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+	if err := deleteFn(deleteCtx, rawRef, auth); err != nil {
```
yocaba, line 203: What if the tag doesn't exist anymore for some reason, would `deleteFn` fail? If so, we should check for existence beforehand, otherwise, the resource will be stuck forever?

pkg/controller/target_controller.go
```diff
@@ -346,6 +363,19 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			// Ensure a RenderArtifact object exists for the pushed OCI artifact, and
+			// create a RenderBinding linking this Target to it.
+			aName := renderArtifactName(target.Namespace, rt.Spec.BaseURL, rt.Spec.Repository, rt.Spec.Tag)
+			bName := renderBindingName(aName, target.Name)
+			if err := r.ensureRenderArtifact(ctx, aName, rt, registry.Spec.Flavor, registryNamespace); err != nil {
+				return ctrl.Result{}, errLogAndWrap(log, err, "failed to ensure RenderArtifact for release")
+			}
+			if err := r.ensureRenderBinding(ctx, target, aName, bName); err != nil {
+				return ctrl.Result{}, errLogAndWrap(log, err, "failed to ensure RenderBinding for release")
+			}
```
yocaba: If the `RenderArtifact` is created and has no `RenderBinding` yet (for a very short amount of time though, but still), the `RenderArtifact` could be garbage-collected in the meantime by the renderartifact_controller?

pkg/controller/renderartifact_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	log.V(1).Info("OCI tag deleted successfully", "ref", rawRef)
+	r.Recorder.Eventf(artifact, nil, corev1.EventTypeNormal,
+		"OCICleanupSucceeded", "Delete",
+		"Successfully deleted OCI tag %s", rawRef)
+
+	return nil
+}
+
+// resolveAuth builds an authn.Authenticator from the artifact's PushSecretRef.
+// Returns authn.Anonymous if no secret is configured or if loading fails.
```
yocaba, line 242: Wondering if that's the expected behavior from a user's perspective. If there are push secrets, I expect them to be mandatory for registry access. Falling back to anonymous would result in an error anyway but later on?

pkg/controller/renderartifact_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		latest := artifact.DeepCopy()
+		apimeta.SetStatusCondition(&latest.Status.Conditions, metav1.Condition{
+			Type:               ConditionTypeOCICleanup,
+			Status:             metav1.ConditionFalse,
+			ObservedGeneration: artifact.Generation,
+			Reason:             "DeleteFailed",
+			Message:            err.Error(),
+		})
+		// Status update, if it fails, the event + log are visible in kubectl
+		if sErr := r.Status().Update(ctx, latest); sErr != nil {
```
yocaba: That's prone for update conflicts. And if the update fails, the Status change is lost since there is no retrying on K8s reconcile level? I guess we might better use Patch.

yocaba (dismissed):
I didn't get through the entire PR yet, but I've added some remarks already that I found also with the help of AI which I already drop here. :)

Will continue reviewing later on.

pkg/controller/renderartifact_controller.go
```diff
@@ -0,0 +1,323 @@
...
+		// Status patch, if it fails, the event + log are visible in kubectl
+		if sErr := r.Status().Patch(ctx, latest, client.MergeFrom(artifact)); sErr != nil {
+			log.Error(sErr, "failed to update status condition after OCI cleanup failure")
+		}
+
+		return err
+	}
+
+	log.V(1).Info("OCI tag deleted successfully", "ref", rawRef)
+	r.Recorder.Eventf(artifact, nil, corev1.EventTypeNormal,
```
trevex, line 234: Do we need to cleanup up the or properly set the `ConditionTypeOCICleanup` as well?

pkg/controller/target_controller.go
```diff
@@ -667,6 +720,125 @@ func (r *TargetReconciler) deleteOwnedRenderTasks(ctx context.Context, target *s
...
+		}
+	}
+
+	return nil
+}
+
+// ensureRenderArtifact creates a RenderArtifact for the given RenderTask's OCI coordinates
+// if one does not already exist. Idempotent: if it already exists (possibly created by
+// another Target reconciling the same shared artifact), this is a no-op.
+func (r *TargetReconciler) ensureRenderArtifact(ctx context.Context, name string, rt *solarv1alpha1.RenderTask, flavor, pushSecretNamespace string) error {
```
trevex, line 868: We should either add a comment to document or pass on the namespace, currently this function assumes secret lives in the same namespaces as RenderTask.

pkg/controller/target_controller.go
```diff
@@ -467,6 +509,17 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
+		// Clean up stale RenderBindings owned by this target.
+		currentBindingNames := map[string]struct{}{bootstrapBindingName: {}}
+		for _, ri := range releases {
+			if ri.artifactBindingName != "" {
+				currentBindingNames[ri.artifactBindingName] = struct{}{}
+			}
+		}
+		if err := r.deleteStaleRenderBindings(ctx, target, currentBindingNames); err != nil {
+			log.Error(err, "failed to clean up stale RenderBindings")
```
trevex, line 606: Is it fine to swallow this error?

pkg/controller/renderartifact_controller_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+// reset clears the recorded calls and removes any configured failure.
+func (s *stubTagDeleter) reset() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.failErr = nil
+	s.calledWith = nil
+}
+
+var _ = Describe("RenderArtifactController", Ordered, func() {
```
trevex, line 67: Should we also add a test covering that a 404 by OCI registry is treated as already deleted correctly?

pkg/controller/renderartifact_controller.go
```diff
@@ -0,0 +1,323 @@
...
+
+	data := secret.Data[corev1.DockerConfigJsonKey]
+	if len(data) == 0 {
+		return authn.Anonymous
+	}
+
+	var cfg struct {
+		Auths map[string]authn.AuthConfig `json:"auths"`
+	}
+	if err := json.Unmarshal(data, &cfg); err != nil {
```
trevex, line 298: We should probably at least log something. A malformed config will just be treates as anonymous...

trevex (dismissed):

pkg/controller/target_controller.go
rebEllieous: It should be, because the if this fails (which should only ever be due to networking errors and such) the next reconcile will retry. And until then this error wont affect anything else.

pkg/controller/renderartifact_controller.go
rebEllieous: We don't as far as i can see (unless you saw something i missed). The finalizer is removed immediately after, the object is deleted, any status write would be lost moments after

pkg/ociregistry/ociregistry_test.go
```diff
@@ -0,0 +1,144 @@
...
+	ref := fmt.Sprintf("%s/ns/repo:v1", host)
+	err := ociregistry.DeleteTag(context.Background(), ref, authn.Anonymous)
+	if err == nil {
+		t.Fatal("expected error on registry failure, got nil")
+	}
+}
+
+// pushMinimalManifest pushes a scratch-level OCI manifest directly via the
+// HTTP API of the test server so that a tag exists for DeleteTag to delete.
+func pushMinimalManifest(srv *httptest.Server, repo, tag string) error {
```
yocaba: Do we need the `pushMinimalManifest` at all? Can we just use: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#Write and https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#ParseReference

Something like:

```
ref, err := name.ParseReference(rawRef)
[..]
if err := remote.Write(ref, empty.Image, remote.WithContext(context.Background())); err != nil {
[..]
}
```

pkg/controller/renderartifact_controller.go
```diff
@@ -0,0 +1,333 @@
...
+		client.InNamespace(artifact.Namespace),
+		client.MatchingFields{indexRenderBindingArtifactName: artifact.Name},
+	); err != nil {
+		return ctrl.Result{}, errLogAndWrap(log, err, "failed to list RenderBindings for RenderArtifact")
+	}
+
+	// If no bindings remain, trigger GC by deleting this object.
+	// The finalizer above will intercept the deletion and handle OCI cleanup.
+	if len(bindingList.Items) == 0 {
+		log.V(1).Info("No RenderBindings remain for RenderArtifact — triggering GC",
```
yocaba, line 152: AI suggested to double-check with calling the API directly in the case there are no bindings found  by using `client.Reader` since 

```
r.List(ctx, bindingList,
		client.InNamespace(artifact.Namespace),
		client.MatchingFields{indexRenderBindingArtifactName: artifact.Name},
	)
```
uses the informer cache. Seems reasonable to me but should be verified if that holds true.

pkg/controller/target_controller.go
```diff
@@ -744,6 +801,129 @@ func (r *TargetReconciler) deleteOwnedRenderTasks(ctx context.Context, target *s
...
+// if one does not already exist. Idempotent: if it already exists (possibly created by
+// another Target reconciling the same shared artifact), this is a no-op.
+//
+// pushSecretNamespace is passed explicitly because the secret may live in a different
+// namespace than the RenderTask (e.g. a cluster-scoped secret namespace chosen by the
+// operator). It must not be inferred from rt.Namespace.
+func (r *TargetReconciler) ensureRenderArtifact(ctx context.Context, name string, rt *solarv1alpha1.RenderTask, flavor, pushSecretNamespace string) error {
+	artifact := &solarv1alpha1.RenderArtifact{}
+	if err := r.Get(ctx, client.ObjectKey{Name: name, Namespace: rt.Namespace}, artifact); err == nil {
+		return nil
```
yocaba, line 877: What happens if `Get` returns a `RenderArtifact` but which has a deletion timestamp set and is about to be deleted?

yocaba (commented):
A few remarks which might be actionable.

trevex (dismissed):

yocaba (approved):
lgtm

PR #541, opened 2026-05-21
feat: finish ReferenceGrants implementation (#490)

dermorz:
## What

Finish the ReferenceGrants feature started in #474.

Closes #490

## Why

PR #474 merged the `ReferenceGrant` CRD and a PoC implementation, but left several gaps:

- The target controller had a bug where cross-namespace `ReleaseBinding` resources (created by the Profile controller in the provider namespace with `spec.targetNamespace` set) were never discovered — the controller only listed bindings in the target's own namespace.
- No documentation existed for operators or users on how to create and manage `ReferenceGrant`s.
- No ADR captured the decision.
- `RegistryBindingSpec` was missing the `TargetNamespace` field needed for future cross-namespace support (blocked on ADR-010).
- Four correctness bugs were discovered and fixed during code review (see Notes for reviewers → Review fixes below).

## Testing

- Unit/integration tests: `make test` — all passing
- `mapReleaseBindingToTarget` and `mapReleaseToTargets` mapper fixes covered by existing reconciler tests
- E2e test extended: the cross-namespace Profile→Target scenario now also verifies that a `RenderTask` is created for the matched target (previously only verified the `ReleaseBinding` was created)
- Integration tests for the cross-namespace ReleaseBinding flow: happy path (grant present → RenderTask created), negative path (no grant → no RenderTask), deduplication (two overlapping grants → exactly one RenderTask), and same-name Target collision (provider-ns Target with identical name must not receive the cross-namespace binding)
- Unit test for `mapReferenceGrantToTargets` verifying the ComponentVersion grant branch enqueues the correct (consumer) Target namespace via a cross-namespace ReleaseBinding

## Notes for reviewers

- **Controller change:** `TargetReconciler.Reconcile` now calls `collectCrossNamespaceReleaseBindings` after the same-namespace list. This lists `ReferenceGrant`s in the target's namespace (already cached) and for each authorized namespace fetches bindings via the existing `indexReleaseBindingTargetName` field index. No new watches added — the `ReferenceGrant` watch was already in place.
- **API change:** `RegistryBindingSpec` gains an optional `TargetNamespace string` field. No controller logic uses it yet (see `FIXME` comment in `target_controller.go`); `make codegen` was run to update generated files.
- **ADR-005** status changed to `superseded`; **ADR-011** added.
- The `ReferenceGrant` user guide page lives in
[...]

PR #570, opened 2026-06-02
fix(controller): silently dropped cross-namespace releases in bootstrap chart

dermorz:
## What
Fix a bug where two same-named Releases from different namespaces silently overwrite each other in the bootstrap chart.

## Why
`buildBootstrapInput` keyed the bootstrap `Releases` map on `ri.name` — the bare Kubernetes Release object name. With cross-namespace ReleaseBindings (#541) two Releases named `my-release` from different namespaces can both survive conflict resolution and target the same Target, colliding on the map key and dropping one deployment silently.

Fix: key on `uniqueName` instead — already guaranteed unique among accepted releases by the resolver. Also extract the derivation rule into a shared `effectiveUniqueName()` helper (was duplicated in both controllers), and add a guard that returns an explicit error if `uniqueName` is empty.

## Testing
- New `buildBootstrapInput` tests: cross-namespace collision (via full `resolveReleaseConflicts` path) and explicit error on empty `uniqueName`.
- `make test` / `make lint` — clean.

## Notes for reviewers
- **Upgrade path:** bootstrap map keys change from the Release object name to `uniqueName`. Existing Targets where the two differ will have inner FluxCD HelmReleases recreated on next bootstrap render.
- `effectiveUniqueName(rel, cv)` lives in `helpers.go`; both controllers use it.
- Docs updated: `rendering-pipeline.md` and ADR-004 now document why `uniqueName` is the bootstrap identity and why the object name is unsuitable.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented release overwrites when releases share the same Kubernetes object name across namespaces by keying bootstrap inputs by a stable per-release unique name.
* **Documentation**
  * Clarified how release unique names are determined and how bootstrap/Helm release names are derived to avoid collisions.
* **Tests**
  * Added tests confirming unique-name enforcement and collision-free bootstrap input generation.

### DIFF
--- a/docs/developer-guide/adrs/004-Unique-Release-Name.md
+++ b/docs/developer-guide/adrs/004-Unique-Release-Name.md
@@ -53,6 +53,7 @@ flowchart LR
     - If priorities are equal, the conflict is broken deterministically by the namespace-qualified ReleaseBinding name in ascending alphabetical order.
 - Releases can define `AntiAffinity` rules (label selectors). If another Release already accepted for a Target matches the selector — or vice versa — the lower-priority Release is excluded.
 - The resolver runs in the Target controller after all ReleaseBindings are collected and before any RenderTask is created.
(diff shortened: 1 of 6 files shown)


dermorz:
@coderabbitai review

yocaba (dismissed):
lgtm

yocaba (approved):

PR #575, opened 2026-06-04
feat: include e2e tests in CI

### LINKED ISSUE #481: Run e2e tests as part of the CI workflow
Currently, our E2E test suite is only executed manually on local developer machines. Because these tests are not integrated into our CI workflow, we lack an automated "safety net" to catch regression bugs before code is merged into the main branch.

Our e2e tests depend on kind.

AC

- e2e tests run as part of our CI workflow

ARC should utilize the solution we've implemented for Solar.

rebEllieous:
## What
Run e2e Tests in CI Pipelines
Closes #481 

## Why
Previously, the E2E test suite had to be triggered manually. To ensure continuous integration and prevent regressions, these tests are now automated within the main pipeline. 
Additionally, the configuration includes adjustments for local execution runners (`act`)

## Testing
- Local Verification: Verified the entire E2E suite locally using `act` with privileged container options and host networking configurations.

## Notes for reviewers
- Local Dev Note: Running the suite locally via `act` now requires an `.actrc` containing `--container-options "--privileged -v /var/run/docker.sock:/var/run/docker.sock"` and `--network host`.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * CI now exposes a reusable image-tag output and a dedicated end-to-end test workflow; Makefile supports configurable registry/tag.

* **Tests**
  * E2E runs consume produced image tags, create image-pull secrets when needed, and include a new values fixture; CI vs local behavior is handled.

* **Bug Fixes**
  * Improved checks to avoid noisy grep output and ensure shellcheck is available before linting.

* **Chores**
  * Local/act runs skip signing, SBOM/provenance, QEMU and GHCR login for faster iteration.


.github/workflows/golang.yaml
```diff
@@ -48,6 +53,11 @@ jobs:
           go-version: ${{ steps.get-go-version.outputs.version }}
       - name: lint-license-and-shell
         run: |
+          # shellcheck is pre-installed on ubuntu-24.04, this is mostly for local execution
+          if ! command -v shellcheck &> /dev/null; then
+            echo "shellcheck is missing. Installing..."
+            sudo apt-get update && sudo apt-get install -y shellcheck
+          fi
```
olzemal, line 53: shellcheck is available with the nix dev shell locally. This would also only work on debian based systems. I would say let's just fail if shellcheck is not available.

.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          # Install kubectl
+          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          chmod +x ./kubectl
+          sudo mv ./kubectl /usr/local/bin/kubectl
+          kubectl version --client
+
+          # Install helm
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+
```
olzemal, line 46: maybe add `helm --version` or similiar as check as well. btw. have you looked into making use of the flake.nix somehow?

.github/workflows/golang.yaml
rebEllieous: Good point. I did most of my local testing using `act`. Relying on a system-wide shellcheck allows us to use the 200MB Ubuntu-lite image instead of the 70GB full image for local runners. While I'm unsure how critical local GitHub Actions testing will be for us going forward, this is a good (or, at elast good enough ) reason for keeping the fallback in place 

You're also right about this currently being limited to Debian-based systems. However, since the actual GitHub workflow is explicitly targeted at Ubuntu, a platform-specific install should be perfectly fine in this context.

Makefile
```diff
@@ -77,37 +83,35 @@ manifests: $(CONTROLLER_GEN) ## Generate ClusterRole and CustomResourceDefinitio
 
 .PHONY: kind-load-local-images
 kind-load-local-images:
-	$(KIND) load docker-image localhost/local/solar-apiserver:$(TAG) --name $(KIND_CLUSTER)
-	$(KIND) load docker-image localhost/local/solar-controller-manager:$(TAG) --name $(KIND_CLUSTER)
-	$(KIND) load docker-image localhost/local/solar-renderer:$(TAG) --name $(KIND_CLUSTER)
-	$(KIND) load docker-image localhost/local/solar-discovery:$(TAG) --name $(KIND_CLUSTER)
-
-KIND_CLUSTER_E2E ?= solar-test-e2e
+	$(KIND) load docker-image $(REGISTRY)/solar-apiserver:$(TAG) --name $(KIND_CLUSTER)
```
yocaba: Nitpick: I guess we can use `APISERVER_IMG` etc. directly here.

test/fixtures/solar-e2e.values.yaml
```diff
@@ -0,0 +1,21 @@
...
+    repository: ghcr.io/opendefensecloud/solar-controller-manager
+
+renderer:
+  image:
+    repository: ghcr.io/opendefensecloud/solar-renderer
+  caConfigMap: root-bundle
+
+discovery:
+  image:
+    repository: ghcr.io/opendefensecloud/solar-discovery-worker
```
yocaba: After we have moved the solar discovery into a standalone tool, it is just solar-discovery. See https://github.com/orgs/opendefensecloud/packages?repo_name=solution-arsenal. Probably something we need to fix somewhere else as well.

Makefile
```diff
@@ -130,10 +134,10 @@ docker-build: docker-build-apiserver docker-build-manager docker-build-discovery
 .PHONY: docker-build-local-images
 docker-build-local-images:
 	$(MAKE) \
-		APISERVER_IMG=localhost/local/solar-apiserver:$(TAG) \
-		MANAGER_IMG=localhost/local/solar-controller-manager:$(TAG) \
-		RENDERER_IMG=localhost/local/solar-renderer:$(TAG) \
-		DISCOVERY_IMG=localhost/local/solar-discovery:$(TAG) docker-build
+		APISERVER_IMG=$(REGISTRY)/solar-apiserver:$(TAG) \
```
yocaba: Why are those variables redefined here?

test/fixtures/solar-e2e.values.yaml
rebEllieous: i've removed the value from both the "base" and the e2e values file, as the image repository for discovery was moved to solar-discovery values ~2 months ago 😅

.github/workflows/test-e2e.yaml
rebEllieous: I added a Helm version check. Could you clarify the intended usage of flake.nix? Do you mean: to run the CI job within flake, or is it primarily meant for extracting specific versions, e.g., for Helm? I considered the latter initially but decided against it, it introduced a lot of complexity for (imo) very little gain

test/e2e/e2e_test.go
```diff
@@ -57,19 +61,53 @@ var _ = Describe("solar", Ordered, func() {
...
+		imageTag = os.Getenv("IMAGE_TAG")
+		imageRepo = os.Getenv("REGISTRY")
+		ghcrToken = os.Getenv("GHCR_TOKEN")
+		ciMode = imageTag != ""
+		if !ciMode {
+			imageTag = "e2e"
+		}
+		if imageRepo == "" {
+			imageRepo = "localhost/local"
+		}
```
yocaba: Wondering if we can keep it simpler and set all env variables in Makefile and make use of them here.

test/e2e/e2e_test.go
```diff
@@ -57,19 +61,53 @@ var _ = Describe("solar", Ordered, func() {
...
-			"--values", filepath.Join(dir, "test", "fixtures", "solar.values.yaml"),
-			"--set", "apiserver.image.tag=e2e",
-			"--set", "controller.image.tag=e2e",
-			"--set", "renderer.image.tag=e2e")
+			"--values", solarValuesFile,
+			"--set", "apiserver.image.tag=" + imageTag,
+			"--set", "controller.image.tag=" + imageTag,
+			"--set", "renderer.image.tag=" + imageTag,
+		}
+		if ciMode && ghcrToken != "" {
```
yocaba: Is it valid to be in CI mode and not have an GHCR token?

.github/workflows/docker.yaml
```diff
@@ -37,19 +42,31 @@ jobs:
...
-      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'ok-to-image')) ||
+      (github.event_name == 'pull_request' && (
+        contains(toJSON(github.event.pull_request.labels), '"ok-to-image"') ||
+        contains(toJSON(github.event.pull_request.labels), '"ok-to-e2e"')
+      )) ||
       (github.event_name == 'release' && github.event.action == 'published')
     runs-on: ubuntu-24.04
+    outputs:
+      image-tag: ${{ steps.meta.outputs.version }}
+      is-local: ${{ env.ACT }}
```
yocaba: Not sure if that ever happens, but `env.ACT` is a string. If someone set it to false, it still evaluates to true.

Makefile
```diff
@@ -82,37 +83,35 @@ manifests: $(CONTROLLER_GEN) ## Generate ClusterRole and CustomResourceDefinitio
...
+	$(KIND) load docker-image $(RENDERER_IMG) --name $(KIND_CLUSTER)
+	$(KIND) load docker-image $(DISCOVERY_IMG) --name $(KIND_CLUSTER)
 
 .PHONY: e2e-cluster
 e2e-cluster: ocm-transfer-demo ## Create a e2e test cluster (Contains everything as a dev-cluster except the solar-api itself)
 	$(MAKE) setup-local-cluster KIND_CLUSTER=$(KIND_CLUSTER_E2E)
-	$(MAKE) docker-build-local-images TAG=e2e
-	$(MAKE) kind-load-local-images TAG=e2e KIND_CLUSTER=$(KIND_CLUSTER_E2E)
-	TAG=e2e KIND_CLUSTER=$(KIND_CLUSTER_E2E) SKIP_SOLAR=true $(HACK_DIR)/dev-cluster.sh
+	@if [ "$(CI)" != "true" ] && [ "$(ACT)" != "true" ]; then \
```
yocaba: Wondering if we can introduce an env variable `E2E_IMAGE_SOURCE` which is set to `local` by default. In the test-e2e.yml we can pass the variable like:

```
make test-e2e E2E_IMAGE_SOURCE=ghcr TAG=$IMAGE_TAG REGISTRY=$REGISTRY
```
and set it to `ghcr` for example. That would help understanding the if condition  here. In addition, we can use it in the test suite and don't have to rely on the IMAGE_TAG not being set like: 
```
ghcrMode = os.Getenv("E2E_IMAGE_SOURCE") == "ghcr"
```

yocaba (changes requested):
Nice that we have the e2e tests running as an CI workflow. :)

All in all, there are a lot of if statements spread over different files that distinguish between local and CI execution which I struggle to understand at first. I added a few suggestion that might help with that a bit, basically saying that we should try to decide between local and CI in the Makefile and pass the env variables to the go test suite based on that.

yocaba (approved):
lgtm

PR #584, opened 2026-06-10
feat: improve resource polling and backoff

### LINKED ISSUE #154: Implement Incremental Backoff for the Creation of RenderTasks, Jobs and Secrets.
**Is your feature request related to a problem? Please describe.**
When there is an error creating a RenderTask, Job or Secret we requeue the reconciliation in a set amount of time (30 seconds).

**Describe the solution you'd like**
The time for the requeue should increment exponentially with the number of retries.

**AC:**
* validate controller runtime behavour to find out whether we should return error or requeue (share finding with team)
* implement backoff for RenderTask, Job and Secret (ReleaseController, HydratedTargetController and RenderTaskController)
* test the behavior (not exhaustively)

AlexHardAtWork:
## What

Closes #154 

## Testing
Unit tests added, E2E green.

## Notes for reviewers
Backing off on errors is a different thing then backing off from resource state polling, so different handling is better for logs.

## Checklist
- [X] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Replaced fixed retry delay with an exponential, age-aware backoff for dependency-wait retries; requeue timing grows with condition age.

* **Bug Fixes**
  * Reconcile paths now use age-aware backoff for missing/denied registries and pending renders, reducing premature retry noise.

* **Tests**
  * Added unit and integration tests validating fresh vs. aged backoff behavior and preservation of condition timestamps during reconcile.

### DIFF
--- a/pkg/controller/backoff.go
+++ b/pkg/controller/backoff.go
@@ -1,0 +1,91 @@
+// Copyright 2026 BWI GmbH and Solution Arsenal contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"time"
+
+	"github.com/cenkalti/backoff/v5"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Dependency-wait backoff curve. 5s initial catches race conditions where the
+// dependency lands seconds later; the 5m cap keeps steady-state polling well
+// below controller-runtime's 1000s rate-limiter cap. Jitter 0.5 avoids
+// thundering herd if many Targets are waiting on the same dependency. No
+// overall cap — we never give up; the dependency may appear hours later when
+// an admin creates it.
+const (
+	dependencyWaitInitialInterval     = 5 * time.Second
+	dependencyWaitMaxInterval         = 5 * time.Minute
+	dependencyWaitMultiplier          = 2.0
+	dependencyWaitRandomizationFactor = 0.5
+)
+
+// requeueAfterForCondition returns the RequeueAfter delay for a dependency-wait
+// condition based on how long it has been in its current Status. A nil or
+// zero-time condition is treated as a fresh wait, returning the initial
+// (jittered) interval.
+//
+// The curve is derived from the condition's LastTransitionTime, which is
+// already managed for us by apimeta.SetStatusCondition (it preserves the
+// timestamp while Status is unchanged and resets it when Status flips). No
+// new persistent state is needed.
+func requeueAfterForCondition(cond *metav1.Condition, now time.Time) time.Duration {
+	age := time.Duration(0)
+	if cond != nil && !cond.LastTransitionTime.IsZero() {
+		if delta := now.Sub(cond.LastTransitionTime.Time); delta > 0 {
+			age = delta
+		}
+	}
+
+	// Walk the backoff curve until the cumulative-time-elapsed-after-this-step
+	// would cross `age`. That step's interval is the right next-RequeueAfter:
+	//   age ∈ [0, I_0)         → return I_0
+	//   age ∈ [I_0, I_0+I_1)   → return I_1
+	//   age ∈ [I_0+I_1, ...)   → return I_2
+	//   ...
+	b := newDependencyWaitBackoff()
+	elapsed := time.Duration(0)
+	// Loop cap is a safety net for nonsensical ages (clock skew, manual
+	// timestamp edits). Even at age=24h with a 5m cap we expect ~290 steps,
+	// well under this bound.
+	for range 1000 {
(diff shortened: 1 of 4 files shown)


pkg/controller/backoff.go
```diff
@@ -0,0 +1,91 @@
...
+	// Loop cap is a safety net for nonsensical ages (clock skew, manual
+	// timestamp edits). Even at age=24h with a 5m cap we expect ~290 steps,
+	// well under this bound.
+	for range 1000 {
+		next := b.NextBackOff()
+		if age < elapsed+next {
+			return next
+		}
+		elapsed += next
+	}
```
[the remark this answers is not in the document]
AlexHardAtWork: backoff/v5's `NextBackOff` already lazy-inits `currentInterval` from
  `InitialInterval` when it's zero
  ([exponential.go:83-86](https://github.com/cenkalti/backoff/blob/v5.0.3/exponential.go#L83-L86)).
  We construct a fresh `*ExponentialBackOff` per call, so the first
  `NextBackOff` invocation initialises correctly without `Reset()`. The
  existing fresh-wait unit test (asserting `[2.5s, 7.5s]` on the first call)
  proves this — it'd fail if `currentInterval=0` produced a wrong value.

  `Reset()` is required when *reusing* an advanced instance, not for a
  fresh literal-constructed one. Skipping.

pkg/controller/backoff.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return dependencyWaitMaxInterval
+}
+
+// requeueAfterForConditions returns the soonest RequeueAfter across the given
+// dependency-wait conditions (min of per-condition values). Nil entries are
+// skipped. Returns 0 if every condition is nil — callers should only invoke
+// this when at least one condition is actually waiting.
+func requeueAfterForConditions(now time.Time, conds ...*metav1.Condition) time.Duration {
```
yocaba: The function is not used. At least not in this PR. Is it used somewhere else? If not, I'd prefer to remove it until we need it.

pkg/controller/backoff_wiring_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	solarv1alpha1 "go.opendefense.cloud/solar/api/solar/v1alpha1"
+)
+
+// These tests cover the wiring of requeueAfterForCondition into TargetReconciler
+// without bringing up envtest. They use controller-runtime's fake client so the
+// suite stays runnable in environments without the kubebuilder etcd binary.
```
yocaba: Just to understand that: Why do we do that since we need envtest for running the other unit test anyway? Is the actual reasoning that we cannot test things with envtest that we need to for this backoff logic?

yocaba (commented):
Nice approach. Just two probably minor thing I've added a comment for.

pkg/controller/backoff_wiring_test.go
AlexHardAtWork: **Why fake client here:** these tests assert on `Reconcile`'s **return value**
  (`ctrl.Result.RequeueAfter`). With envtest, the manager loop consumes the
  return — it goes into the workqueue's internal `AddAfter` and there's no
  public API to read it back. 
  
  Fake client lets us invoke `Reconcile` synchronously and inspect the
  `Result` directly, which is exactly what we want to test.

  **On the "we need envtest anyway" point:** that's true for the existing
  integration tests in `suite_test.go` — they assert on cluster-state
  convergence over time, which absolutely needs a real API server. 
  
  The backoff wiring isn't testing convergence; it's testing the function-shape
  of a single `Reconcile` call. Different concern, different tool.

pkg/controller/backoff.go
AlexHardAtWork: It was thought for rare edge-cases, YAGNI, removed.

pkg/controller/backoff_wiring_test.go
yocaba: I see. But strictly speaking that is not what the comment says, is it. It says: 

> They use controller-runtime's fake client so the suite stays runnable in environments without the kubebuilder etcd binary.

At least that confused me. But it might be just me. :)

AlexHardAtWork: That sounds like confusion is a bad thing now, comment improved ;)

yocaba: nice. thx. :)

yocaba (approved):
lgtm

PR #590, opened 2026-06-11
fix: fix empty resourceversion on multiple componentversions

rebEllieous:
## What
bugfix: multiple componentversions for a component led to empty resourceversion which errored on component update, therefore not creating the second..nth compoentverison for a component

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when creating or updating components and their versions so existing resources are correctly updated rather than causing creation errors.

* **Tests**
  * Added tests to verify conflict-free update behavior when target resources already exist, ensuring specs are populated and no errors are emitted.


dermorz:
Let's add a regressen test and we are good to go. :+1:

dermorz (approved):
Thanks for adding the tests.

LGTM 🚀

PR #612, opened 2026-06-16
chore: update release checksum

rebEllieous:
## What
Fixes the checksum step in release

go build outputs directories inside bin, which cases sha256sum to fail. we find and hash regular files only

## Summary by CodeRabbit

* **Chores**
  * Updated the release workflow's checksum generation process to compute SHA-256 hashes with deterministic ordering, improving the reproducibility and reliability of generated checksums.


dermorz (approved):
lgtm 🚀

PR #617, opened 2026-06-17
feat: implemented e2e test for strict mode registry binding

### LINKED ISSUE #568: Add E2E test for --registry-binding-strict mode
Background:
  The controller manager supports a --registry-binding-strict flag (added in item 5 of #440). In strict
  mode, resolveResources() returns an error when a resource's registry host has no matching
  RegistryBinding, rather than falling through to anonymous pull (relaxed mode, the default). The flag is
   wired through the Helm chart via controller.args.registryBindingStrict. Unit test coverage for strict
  mode exists in pkg/controller/target_controller_test.go, but there is no E2E-level validation.

  What's needed:
  Add an E2E scenario that:
  1. Starts the controller manager with --registry-binding-strict enabled.
  2. Creates a Target with a ReleaseBinding whose ComponentVersion resources reference a registry host
  that has no matching RegistryBinding.
  3. Verifies the Target receives a MissingRegistryBinding status condition.
  4. Verifies that rendering is blocked — no RenderTask is created for the affected release.

  Acceptance criteria:
  - E2E test passes in CI with strict mode enabled for the scenario.
  - The test confirms that adding the missing RegistryBinding (and triggering reconciliation) clears the
  condition and allows rendering to proceed.

_Originally posted by @trevex in https://github.com/opendefensecloud/solution-arsenal/pull/553#discussion_r3332139685_

rebEllieous:
## What
Adds E2E coverage for the --registry-binding-strict controller-manager flag
Closes #568 

## Why
The strict-mode logic in resolveResources had unit and envtest coverage but no end-to-end validation.

## Testing
`make test-e2e`
relaxed mode creates RenderTasks even if no RegistryBinding exists, strict mode does not unless / until a binding exists

## Notes for reviewers
-

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **Tests**
  * Enhanced end-to-end test coverage for strict registry-binding mode, including validation of rendering behavior with and without required registry bindings.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -31,6 +31,7 @@ var _ = Describe("solar", Ordered, func() {
 	var imageRepo string
 	var ciMode bool
 	var ghcrToken string
+	var solarValuesFile string
 	testStart := time.Now()
 
 	SetDefaultEventuallyTimeout(10 * time.Minute)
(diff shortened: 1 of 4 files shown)


dermorz:
@coderabbitai review

rebEllieous:
@coderabbitai review

test/e2e/e2e_test.go
yocaba: We don't wait for the API server to become available for the second redeploy. I guess we better should?

yocaba: We need to update it after the second redeploy so that log fetching works in AfterAll.

test/e2e/e2e_test.go
```diff
@@ -1296,5 +1314,194 @@ var _ = Describe("solar", Ordered, func() {
...
+				}).Should(Succeed())
+
+				By("verifying MissingRegistryBinding condition is never set in relaxed mode")
+				Consistently(func(g Gomega) {
+					cmd := exec.Command(kubectlBinary, "get", "target", "relax-tgt",
+						"-n", testns,
+						"-o", `jsonpath={.status.conditions[?(@.type=="ReleasesRendered")].reason}`)
+					output, err := run(cmd)
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(output).NotTo(Equal("MissingRegistryBinding"),
```
yocaba: Should we also check that the condition equals some success state to ensure that rendering actually succeeded?

test/e2e/e2e_test.go
```diff
@@ -1296,5 +1314,194 @@ var _ = Describe("solar", Ordered, func() {
...
+					{"registrybinding", "strict-rb-source"},
+					{"registry", "strict-source-registry"},
+				} {
+					cmd := exec.Command(kubectlBinary, "delete", res[0], res[1],
+						"-n", testns, "--ignore-not-found", "--timeout=2m")
+					_, _ = run(cmd)
+				}
+
+				By("restoring controller-manager to relaxed registry binding mode")
+				redeploySolar()
```
yocaba, line 1404: Is the restoring needed since we don't run any other tests afterwards?

yocaba (changes requested):
lgtm overall, some minor remarks for potential minor improvements

test/e2e/e2e_test.go
rebEllieous: we dont "need" it, but it's not a bad place to have it. We could also burden the next test (one we have one) to redeploy solar with the cnfigs it needs, but it makes sense to "revert to the default" within our specialized context.
the main reason i've included it is for future merges and such, we wouldnt want any new tests being merged to fail or to have to know to redeploy solar :)

yocaba (approved):
lgtm

PR #621, opened 2026-06-17
feat(controller): implement deletion protection via controller-managed finalizers

### LINKED ISSUE #308: Deletion protection
As pointed out by @dermorz when a Release references a ComponentVersion we probably don't want allow the deletion of the ComponentVersion. This is specifically for "shared" ComponentVersions (e.g. Cluster or ReferenceGranted) relevant as if something is provided for other we don't want to negatively impact their deployments.

dermorz:
## What
Implements deletion protection for `ComponentVersion`, `Release`, `Registry`, and `ReleaseBinding` resources using controller-managed finalizers, preventing cascade-deletion accidents when upstream resources are deleted while still referenced.

Closes #308 

## Why
Before this change, deleting a `Component` while `ComponentVersions` still pointed to it, or a `Release` while active deployments existed, would cause dangling references and silent failures downstream. This adds a guard at the API level: referenced resources cannot be deleted until all their referencers are gone.

## Testing
All 17 envtest/Ginkgo suites pass. Each new controller is covered by dedicated tests:
- Add-finalizer-on-create (both self-finalizer and protection finalizer)
- Block-deletion-while-referenced (Consistently assert DeletionTimestamp present, resource not gone)
- Remove-protection-finalizer-when-last-reference-deleted
- Retain-protection-finalizer-when-a-second-referencer-still-exists (multi-referencer scenario)
- Profile-owned-binding guard: `removeReleaseRefFinalizer` deferred to Profile controller while `profileFinalizer` is present, verified with a test-only blocker finalizer

`make test-e2e` timeout bumped from the implicit 10 m Go default to 15 m; new controllers add reconcile cycles to namespace teardown (previously resources were GC'd immediately).

## Notes for reviewers

**New controllers** — `ComponentVersionReconciler`, `ReleaseBindingReconciler`, and `RegistryBindingReconciler` are registered in `solar-controller-manager`. During a rolling upgrade, the brief window between old pod termination and new pod startup leaves existing resources unprotected; this is unavoidable with a rolling deploy and the window is short.

**RBAC changes** — `patch`/`update` verbs added to `components`, `componentversions`, `profiles`, `registries`, and `registrybindings`; new `/finalizers` sub-resource rules added for all of those plus `releases` and `releasebindings`. See `charts/solar/files/role.yaml`.

**New finalizers on existing resources** — after upgrade, all existing `ComponentVersions`, `ReleaseBindings`, `RegistryBindings`, and `Releases` will have finalizers added on their next reconcile. Resources already in Terminating state before upgrade will be unblocked normally once the new controllers are running.

**No API schema changes** — only an informational doc comment added to `Profile` describing cascade-deletion behaviour.

**Finalizer placement** — all protection-finalizer patches use a fresh `Get` immediately before the `Patch` to avoid a JSON-merge-patch race where a concurrent add could be overwritten by a stale base object (RFC 7396: only fields present in the patch body are applied; `resourceVersion` is excluded from the diff when both base and modified share the same value).

## Checklist
- [x] Tests added/updated

## Release Notes

* **New Features**
  * Added ComponentVersion, ReleaseBinding, and RegistryBinding controllers with coordinated finalizer-based deletion protection and safer cascade cleanup across related resources.
* **Documentation**
  * Added developer guides for the new controllers.
  * Expanded Profile deletion documentation (developer guide and API reference) with clearer destructive/cascading and undeployment-avoidance guidance.
* **Chores / Tests**
  * Updated RBAC permissions to support finalizer and reference handling.
  * Increased e2e test timeout for improved stability.

### DIFF
--- a/pkg/controller/helpers.go
+++ b/pkg/controller/helpers.go
@@ -8,9 +8,11 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -37,8 +39,33 @@ const (
 	// Field index key for looking up RenderBindings by the RenderArtifact they reference.
 	indexRenderBindingArtifactName = "spec.renderArtifactRef.name"
 
+	// Field index keys for deletion-protection reference lookups.
+	// Release: composite "<cvNamespace>/<cvName>" resolving cross-namespace refs.
+	indexReleaseByCVRef = "dp.spec.componentVersionRef"
+	// ComponentVersion: same-namespace lookup by component name.
+	indexCVByComponentName = "dp.spec.componentRef.name"
+	// Profile: same-namespace lookup by release name.
+	indexProfileByReleaseName = "dp.spec.releaseRef.name"
+	// Target: composite "<registryNamespace>/<registryName>" resolving cross-namespace refs.
+	indexTargetByRegistryRef = "dp.spec.renderRegistryRef"
+	// RegistryBinding: same-namespace lookup by registry name.
+	indexRegistryBindingByRegistryName = "dp.spec.registryRef.name"
+
 	maxK8sObjectNameLen = 253
 	maxK8sLabelValueLen = 63
+
+	// Self-finalizers: added to the referencing resource so the controller can observe deletion.
+	releaseFinalizer          = "solar.opendefense.cloud/release-finalizer"
+	profileFinalizer          = "solar.opendefense.cloud/profile-finalizer"
+	releaseBindingFinalizer   = "solar.opendefense.cloud/releasebinding-finalizer"
+	componentVersionFinalizer = "solar.opendefense.cloud/componentversion-finalizer"
+	registryBindingFinalizer  = "solar.opendefense.cloud/registrybinding-finalizer"
+
+	// Protection finalizers: added to the referenced resource to block deletion while referenced.
+	componentVersionRefFinalizer = "solar.opendefense.cloud/componentversion-ref"
+	componentRefFinalizer        = "solar.opendefense.cloud/component-ref"
+	releaseRefFinalizer          = "solar.opendefense.cloud/release-ref"
+	registryRefFinalizer         = "solar.opendefense.cloud/registry-ref"
 )
 
 // truncateName truncates a name to maxLen characters. If truncation is needed,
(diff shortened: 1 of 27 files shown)


pkg/controller/suite_test.go
```diff
@@ -283,11 +310,86 @@ var _ = AfterEach(func() {
...
+	}
+
+	registryList := &solarv1alpha1.RegistryList{}
+	if err := k8sClient.List(ctx, registryList, client.InNamespace(ns.Name)); err == nil {
+		objs := make([]client.Object, len(registryList.Items))
+		for i := range registryList.Items {
+			objs[i] = &registryList.Items[i]
+		}
+		forceDelete(objs)
+	}
```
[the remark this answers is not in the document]
dermorz: Dropping errors in cleanup is intentional to avoid aborting teardown halfway through if one resource fails.

pkg/controller/profile_controller_test.go
```diff
@@ -164,6 +167,209 @@ var _ = Describe("ProfileReconciler", Ordered, func() {
...
+				g.Expect(updated.DeletionTimestamp).NotTo(BeNil())
+			}, consistentlyDuration).Should(Succeed())
+
+			// Delete Profile — controller removes releaseRefFinalizer from Release,
+			// then removes profileFinalizer, unblocking Release deletion.
+			Expect(client.IgnoreNotFound(k8sClient.Delete(ctx, profile))).To(Succeed())
+
+			Eventually(func() error {
+				return client.IgnoreNotFound(k8sClient.Get(ctx, client.ObjectKeyFromObject(release), &solarv1alpha1.Release{}))
+			}, eventuallyTimeout).Should(Succeed())
```
trevex: Small but meaningful test gap: `client.IgnoreNotFound(Get(...))` returns `nil` both when the object is gone **and** when the Get succeeds (object still present), so `Should(Succeed())` passes on the very first poll — before the Release is actually deleted. That means the *unblock-after-delete* half of this "blocks Release deletion" test isn't really being verified.

This is the same pitfall the earlier commit fixed elsewhere — `releasebinding_controller_test.go:106` has the correct shape:

```go
Eventually(func() bool {
    err := k8sClient.Get(ctx, client.ObjectKeyFromObject(release), &solarv1alpha1.Release{})
    return apierrors.IsNotFound(err)
}, eventuallyTimeout).Should(BeTrue())
```

Mind switching to that form here? It's a quick change that closes a real verification gap.

trevex (commented):
Did a focused pass on the finalizer/reconcile correctness here — really nice work overall. The cleanup ordering (external finalizer first, self-finalizer last), the single-cycle protection fix, and especially the Profile→ReleaseBinding cascade with the blocker-finalizer test are all solid, and the RBAC additions line up with the controller markers. 🙌

A couple of things I'd love your take on before merge — left two test/correctness items + a couple of smaller notes inline. None are blockers in the production happy path, but the two test assertions and the optimistic-lock note feel worth addressing.

One extra consistency nit that didn't land on a diff line (the code is unchanged here, `target_controller.go` around L168-181): the Target controller still uses the two-cycle pattern — add the self-finalizer, then `return ctrl.Result{}, nil` — so the Registry `registry-ref` protection lands on the *next* reconcile rather than the same one, unlike the four controllers the recent fix collapsed into a single cycle. The requeue is immediate so the window is tiny, but it's a slightly wider protection gap than the others. Worth aligning for consistency, or a brief comment on why Target is intentionally different? (Also flagging since the feature commit said "all 5 controllers" but Target wasn't in the changed set.)

pkg/controller/profile_controller_test.go
```diff
@@ -164,6 +167,209 @@ var _ = Describe("ProfileReconciler", Ordered, func() {
...
+			Expect(k8sClient.Get(ctx, bindingKey, bindingDeleting)).To(Succeed())
+			bindingWithoutBlocker := bindingDeleting.DeepCopy()
+			bindingWithoutBlocker.Finalizers = slices.DeleteFunc(bindingWithoutBlocker.Finalizers,
+				func(s string) bool { return s == testBlocker })
+			Expect(k8sClient.Patch(ctx, bindingWithoutBlocker, client.MergeFrom(bindingDeleting))).To(Succeed())
+
+			// Profile must eventually complete deletion (fully removed from API).
+			Eventually(func() error {
+				return client.IgnoreNotFound(k8sClient.Get(ctx, client.ObjectKeyFromObject(profile), &solarv1alpha1.Profile{}))
+			}, eventuallyTimeout).Should(Succeed())
```
trevex: Same `IgnoreNotFound`-as-deletion-check issue as above — this assertion passes on the first poll regardless of whether the Profile is actually gone, so it's effectively a no-op. Suggest the `apierrors.IsNotFound(...)).Should(BeTrue())` form so we genuinely assert the Profile completes deletion. (The `release-ref` removal assertion just below it looks correctly written, for what it's worth.)

pkg/controller/registrybinding_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	if rb.Spec.RegistryRef.Name != "" {
+		registry := &solarv1alpha1.Registry{}
+		if err := r.Get(ctx, types.NamespacedName{Name: rb.Spec.RegistryRef.Name, Namespace: rb.Namespace}, registry); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return ctrl.Result{}, errLogAndWrap(log, err, "failed to get Registry for protection finalizer")
+			}
+		} else if !slices.Contains(registry.Finalizers, registryRefFinalizer) {
+			latest := registry.DeepCopy()
+			latest.Finalizers = append(latest.Finalizers, registryRefFinalizer)
+			if err := r.Patch(ctx, latest, client.MergeFrom(registry)); err != nil {
```
trevex: Not a blocker, but worth a thought on concurrency: the Registry's `registry-ref` finalizer is patched by both this controller and the Target controller, and Release's `release-ref` similarly by Profile + ReleaseBinding. A plain `client.MergeFrom` JSON-merge-patch replaces the whole `finalizers` array, so two controllers acting on stale reads can drop each other's additions/removals — on the removal path that could silently strip a still-needed protection finalizer.

Would `client.MergeFromWithOptimisticLock(...)` on these shared-object finalizer add/remove patches be reasonable? That way a conflicting write 409s and we just requeue, instead of a lost update. (Applies to the matching patch sites in `target_controller.go`, `profile_controller.go`, and `helpers.go` too.)

pkg/controller/helpers.go
```diff
@@ -347,3 +450,67 @@ func indexRenderBindingFields(ctx context.Context, mgr ctrl.Manager) error {
...
+	if err := c.Get(ctx, client.ObjectKeyFromObject(registry), freshRegistry); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+
+		return errLogAndWrap(ctrl.LoggerFrom(ctx), err, "failed to get latest Registry for finalizer removal")
+	}
+	original := freshRegistry.DeepCopy()
+	freshRegistry.Finalizers = slices.DeleteFunc(freshRegistry.Finalizers, func(s string) bool { return s == registryRefFinalizer })
+	if err := c.Patch(ctx, freshRegistry, client.MergeFrom(original)); err != nil {
```
trevex: Minor / mostly a doc-comment suggestion: the reference count here comes from the cached client List, so there's a small stale-read window where a just-`Create`d referencer isn't in the informer cache yet and we'd remove the protection finalizer while a live referencer exists. The referencer's own reconcile re-adds protection, so the window is tiny and largely unavoidable without an uncached read — but a short comment documenting the accepted race would help future readers not mistake it for a bug.

trevex (approved):
looks good 👍

PR #622, opened 2026-06-18
fix: make deleting OCI tags work in CI e2e tests

### LINKED ISSUE #596: deleteStaleRenderBindings never GCs stale RenderBinding in CI
**Describe the bug**
The e2e test `RenderArtifact and RenderBinding lifecycle > should retain the RenderArtifact when only one binding is removed` consistently fails in CI but passes locally. The target controller reconciles after `art-rb-1` is deleted but never cleans up `art-tgt-1`'s stale RenderBinding, even after the full 10-minute timeout.

**To Reproduce**
make test-e2e E2E_IMAGE_SOURCE=ghcr TAG=pr-575 REGISTRY=ghcr.io/opendefensecloud

yocaba:
## What
Closes #596.

## Why
e2e tests related to garbage-collecting OCI tags in the render registry has been skipped, because they failed when running within a GitHub action. The reason was a missing CA certificate when deleting an OCI tag with the help of go-containerregistry in the controller process (RenderArtifact reconcilation).

Pushing rendered artifacts to the very same render registry worked, because pushing is done in a separate RenderTask pod where we mount the CA. Locally it has worked, because we use another values file where CA certificate mounting has been enabled.

## Testing
`make test-e2e` locally and e2e tests as a check in this PR.

## Checklist
- [x] Tests added/updated

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -27,6 +27,7 @@ var _ = Describe("solar", Ordered, func() {
 	var controllerPodName string
 	var testns string
 	var deployns string
+	var registryns string
 	var imageTag string
 	var imageRepo string
 	var ciMode bool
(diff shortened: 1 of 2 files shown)


dermorz (approved):
lgtm 🚀

PR #660, opened 2026-06-26
fix: handle insecure deploy registry for bootstrap chart

### LINKED ISSUE #631: Renderer omits insecure: true from generated OCIRepository manifests when deploy registry is plain HTTP
When Registry.spec.plainHTTP: true, the bootstrap chart the renderer produces contains OCIRepository resources without spec.insecure: true. Flux therefore tries HTTPS against the plain-HTTP registry, fails silently, and every child HelmRelease stays InProgress indefinitely.

ACs:

- [ ]  OCIRepository resources in the rendered bootstrap chart include spec.insecure: true when the deploy registry has plainHTTP: true

- [ ]  HTTPS clusters are unaffected (spec.insecure absent or false)

olzemal:
## What

Closes #631

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Bootstrap and registry handling now supports plain HTTP connections, improving compatibility with insecure registries.

* **Bug Fixes**
  * Updated bootstrap comparison logic so changes to registry connection mode correctly trigger a refresh when needed.

### DIFF
--- a/pkg/controller/target_controller.go
+++ b/pkg/controller/target_controller.go
@@ -537,7 +537,7 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	default:
 		// RenderTask exists — check if the desired bootstrap input changed
 		// (release set, resolved refs/tags, or userdata)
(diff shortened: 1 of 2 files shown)


yocaba (approved):
lgtm

PR #678, opened 2026-07-05
fix(hooks): improve pre-commit hook performance (#669)

### LINKED ISSUE #669: Investigate and improve performance of pre-commit hooks
**Describe the bug**
The fmt and lint commit hooks can take a long time. We should evaluate why the hooks are taking long sometimes and improve it.

**To Reproduce**
Change (go) code and commit.

**Expected behavior**
Commit process is reasonable short.

Gravirei:
## What
Improve performance of pre-commit hooks by eliminating redundant lint and format runs.

Closes #669

## Why
The fmt and lint pre-commit hooks were slow because:
1. The `golangci-lint` hook ran `make golangci-lint` (full-project scan) before also linting each changed directory individually — a redundant double scan.
2. The `gofmt` hook was redundant since `golangci-lint` already invokes `gofmt` as a formatter.

## Testing
Pre-commit hook invocation can be verified via `devenv shell` by staging Go files and attempting a commit.

## Notes for reviewers
Configuration change only — no code changes.

## Checklist
- [ ] Tests added/updated

## Summary by CodeRabbit

* **Chores**
  * Improved the Go lint hook so it runs directly on the affected directories, making checks faster and more targeted.
  * Updated development configuration to rely on the existing Git hook setup and removed automatic Go formatting enablement from language settings.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -27,8 +27,29 @@
     text = ''
         #!/usr/bin/env bash
         set -e
+
+        BINDIR="$(cd "$(dirname "$0")" && pwd)"
+        GOLANGCI_LINT="$BINDIR/golangci-lint"
+
+        if [ ! -x "$GOLANGCI_LINT" ]; then
+          if command -v go &>/dev/null && [ -f "$BINDIR/../tools.lock" ]; then
+            TOOLSPEC=$(grep '^golangci-lint ' "$BINDIR/../tools.lock" | head -1)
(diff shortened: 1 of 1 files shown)


yocaba:
Thanks for digging into #669 — dropping the redundant make golangci-lint full scan is a genuine speedup. But this PR bundles multiple things that are not all about speed and each has tradeoffs.

We'll track the redundant-scan fix on #669 directly. Closing this PR.

PR #700, opened 2026-07-13
fix: registry resources

### LINKED ISSUE #623: solar-discovery chart does not create Registry resources
**Describe the bug**
After deploying solar-discovery there is a message showing the configured registries. However the values for this are only ever used for this message and no Registry CR gets ever created.

https://github.com/opendefensecloud/solution-arsenal/blob/8d1cbd4b7495c9828a35afd8e6590b66abea4f4b/charts/solar-discovery/templates/NOTES.txt#L8

**Expected behavior**
Either the function of registering a registry directly via helm is removed or fully implemented.

AlexHardAtWork:
## What
Closes #623

## Testing
E2e tested.

## Checklist
- [X] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added Helm chart support for creating and managing Registry CRs via a new `registries` values section (with per-registry webhook/scan, secrets, HTTP, flavor, and scan interval).
  * Chart-managed Registries are deleted on `helm uninstall`.
* **Documentation**
  * Expanded guidance for Registry lifecycle (embedded vs standalone) and clarified behavior when `registries` is empty, including multi-tenant/shared setups.
* **Tests**
  * Added E2E coverage for Helm ownership/lifecycle when `registries` is unset or provided, including uninstall deletion and a validation check when a registry entry omits `hostname`.


cbrgm:
Found a small bug in `templates/registries.yaml` (or more a less a small UX nitpick 😄)

Line 2 builds the name with `lower .hostname`, before line 8 checks `required "registries[].hostname is required" .hostname`. So if `hostname` is missing (key not set), rendering fails already at line 2, not at the `required` check:

```
Error: ... at <.hostname>: wrong type for value; expected string; got interface {}
```

Afaik the `required` message on line 8 is basically dead code for this case. It only fires when `hostname` is set but empty string, not when the key is absent.

Repro, no cluster needed:
```yaml
# values.yaml
registries:
  - name: my-registry
```
```
helm template test ./charts/solar-discovery -f values.yaml --show-only templates/registries.yaml
```

Install still fails either way, so no silent corruption, just a confusing error instead of the intended one. Simple fix imo, guard `.hostname` with `required` before using it for the name derivation on line 2

cbrgm:
Otherwise looks good to me. I also ran the E2E test locally and seems to work just fine. The registries are also being created

charts/solar-discovery/templates/registries.yaml
```diff
@@ -0,0 +1,32 @@
+{{- range .Values.registries }}
+{{- $name := .name | default (regexReplaceAll "[^a-z0-9-]" (lower .hostname) "-") -}}
```
cbrgm: Suggested fix, guard `.hostname` with `required` here, before it gets used in the name derivation.

```suggestion
{{- $hostname := required "registries[].hostname is required" .hostname }}
{{- $name := .name | default (regexReplaceAll "[^a-z0-9-]" (lower $hostname) "-") -}}
```

charts/solar-discovery/templates/registries.yaml
```diff
@@ -0,0 +1,32 @@
...
+---
+apiVersion: solar.opendefense.cloud/v1alpha1
+kind: Registry
+metadata:
+  name: {{ required "registries[].hostname is required" $name }}
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "solar-discovery.labels" $ | nindent 4 }}
+spec:
+  hostname: {{ required "registries[].hostname is required" .hostname | quote }}
```
cbrgm: Follow up to the suggestion on line 2, reuse `$hostname` here instead of checking `.hostname` again.

```suggestion
  hostname: {{ $hostname | quote }}
```

charts/solar-discovery/templates/registries.yaml
AlexHardAtWork: good catch, fixed, test added.

charts/solar-discovery/templates/registries.yaml
AlexHardAtWork: good catch, fixed, test added.

cbrgm (approved):

PR #704, opened 2026-07-15
feat(solar-agent): distribute signing public key via solar-agent (#690)

### LINKED ISSUE #690: Distribute Public Key via solar-agent
**Description**

Extend the solar-agent to read the public key from the Target's status (or a referenced Secret in
the Solar cluster) and write it to the target cluster so that Flux
`OCIRepository` resources can reference it in `spec.verify`.

**Alternatively**: Targets expect a keypair to be generated by the cluster-operator when registering the cluster. (scratch this issue, no automatic keypair retrieval)

**Acceptance Criteria**

- [ ] The `solar-agent` authenticates to the Solar API and retrieves the
      public key for its assigned Target.
- [ ] The agent creates or updates a Kubernetes Secret on the target cluster
      containing the public key (e.g., `cosign.pub`).
- [ ] The agent monitors the Target for key rotation events and updates the
      Secret accordingly.
- [ ] During the rotation grace period, the agent maintains both old and new
      public keys so that verification does not break mid-rollout.
- [ ] When the Target is deleted, the agent removes the public key Secret
      from the target cluster.
- [ ] Agent-level unit tests cover: initial distribution, key rotation,
      target deletion, connection failure.

**Additional Context**

This item is blocked on the existence of the `solar-agent`.

**Follow-up to:** [solution-arsenal#655](https://github.com/opendefensecloud/solution-arsenal/issues/655)

Gravirei:
## What
Extend the target_controller to generate ECDSA P-256 keypairs and create the solar-agent to distribute the public key as a Kubernetes Secret.

Closes #690

## Why
This enables Flux OCIRepository verification on target clusters. Each target gets its own keypair; the public key is distributed by the solar-agent as a cosign-pub Secret in the Target's namespace.

## Testing
- Unit tests in pkg/agent/ (6 tests): initial distribution, skip-without-key, key rotation, target deletion, non-existent target, no-op on matching secret
- All existing controller unit tests pass
- go vet and go build clean

## Notes for reviewers
- Two new fields added to TargetStatus: PublicKey and PreviousPublicKey (deepcopy not needed — value types)
- New binary: cmd/solar-agent (controller-runtime based)
- New package: pkg/agent/ (Reconciler watching Targets)
- Key generation uses standard crypto/ecdsa with P-256 (cosign-compatible)

## Summary by CodeRabbit

- **New Features**
  - Targets now automatically receive signing keys when needed.
  - Public verification keys are distributed to target clusters through Kubernetes Secrets.
  - Key rotation preserves the previous public key temporarily to support in-flight verification.
  - Added status reporting for verification key distribution.
  - Added the Solar agent controller with health and readiness monitoring.

- **Bug Fixes**
  - Ensured verification Secrets are updated, retained, and removed appropriately as targets change or are deleted.


yocaba:
The key distribution is blocked upstream on the solar-agent (which doesn't exist yet), so it can't move forward. Closing.

PR #706, opened 2026-07-16
test: more target controller testing

### LINKED ISSUE #604: test: cover target_controller.go cross-namespace Registry/ReferenceGrant paths
## Background

`pkg/controller` overall sits at 75.9% (944/1243 statements), with `target_controller.go` as the largest remaining contributor. This issue covers the cross-namespace Registry/ReferenceGrant coverage gap in that file.

## Scope

Two functions are completely uncovered.

- `registryGranted` (line 1089, 0%) - checks whether a `ReferenceGrant` in the registry's namespace permits a `Target` in another namespace to reference it.
- `grantPermitsRegistryAccess` (line 1106, 0%) - the `Registry`-specific wrapper around `grantPermits`.

Related, partially-covered functions that the same test scenarios would help close.

- `Reconcile` (63.9%) - specifically the `target.Spec.RenderRegistryNamespace != ""` branch (lines ~167-216), covering granted/denied/registry-not-found/missing-`SolarSecretRef` outcomes for `ConditionTypeRegistryResolved`.
- `mapRegistryToTargets` (50.0%) - the cross-namespace `ReferenceGrant` branch (lines ~1141-1180).
- `mapReferenceGrantToTargets` (59.5%) and `grantsRegistryResource` (75.0%) - the `Registry`-resource-kind branch.

`pkg/controller/target_controller_test.go` already has the building blocks for this: `newReferenceGrant` (line 1163) and the cross-namespace `ReleaseBinding` + `ReferenceGrant` scenarios (lines 1177-1252) follow exactly the pattern needed here, just for `Registry` instead of `Release`/`ReleaseBinding`.

Suggested new specs, mirroring the existing `RegistryResolved=False` specs at lines 123-153.

- A `Target` with `RenderRegistryNamespace` set to a different namespace and no `ReferenceGrant` should produce `RegistryResolved=False` with reason `NotGranted`.
- The same setup plus a `ReferenceGrant` permitting `Target -> Registry` should produce `RegistryResolved=True` with reason `Resolved`.
- A `Registry` create/update event in a namespace with a granting `ReferenceGrant` should cause `mapRegistryToTargets` to enqueue the cross-namespace `Target`.

## References

- `pkg/controller/target_controller_test.go` lines 123-153, 1163-1252, 1401-1434

AlexHardAtWork:
## What
Closes #604 

## Checklist
- [X] Tests added/updated

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of cross-namespace registry access permissions for targets.
  - `ConditionTypeRegistryResolved` now accurately reflects `Resolved` vs `NotGranted`, including failure scenarios when permission lookups error.
  - Target reconciliation is now triggered appropriately when cross-namespace `ReferenceGrant` rules change.

- **Tests**
  - Added coverage for registry grant matching and permission lookup error propagation.
  - Added specifications verifying cross-namespace registry resolution and correct reconciliation requests for `Target` updates.


yocaba:
Overall lgtm. Do we want to cover the mentioned paths in the ticket as well or are they left out on purpose:

> not-found/missing-SolarSecretRef outcomes for ConditionTypeRegistryResolved.

pkg/controller/target_controller_grants_test.go
```diff
@@ -110,4 +110,41 @@ func TestRegistryGranted(t *testing.T) {
 			t.Error("expected registry access to be denied")
 		}
 	})
+
+	t.Run("returns an error when the List call fails", func(t *testing.T) {
+		t.Parallel()
+		// Scheme without ReferenceGrant registered so listing fails.
+		bareScheme := runtime.NewScheme()
+		_ = scheme.AddToScheme(bareScheme)
```
cbrgm: Small thing: AFAIK the core scheme is not really needed here, the List already fails because the solar types are not registered. So we could drop this line, or better expand the comment to say why we still add it

pkg/controller/target_controller_grants_test.go
```diff
@@ -110,4 +110,41 @@ func TestRegistryGranted(t *testing.T) {
 			t.Error("expected registry access to be denied")
 		}
 	})
+
+	t.Run("returns an error when the List call fails", func(t *testing.T) {
+		t.Parallel()
+		// Scheme without ReferenceGrant registered so listing fails.
+		bareScheme := runtime.NewScheme()
```
cbrgm: Nit: here we name it `bareScheme` but the outer test uses `sch`  (see line 66 for example). Not wrong, only a bit inconsistent, IDK if worth to change. Otherwise LGTM

pkg/controller/target_controller_grants_test.go
```diff
@@ -110,4 +110,41 @@ func TestRegistryGranted(t *testing.T) {
...
+		}
+	})
+}
+
+func TestGrantsRegistryResource(t *testing.T) {
+	t.Parallel()
+
+	t.Run("permits when To lists a Registry resource", func(t *testing.T) {
+		t.Parallel()
+		grant := newRegistryGrant("target-ns", "Registry")
```
cbrgm, line 143: `grantsRegistryResource` only looks at the `To` list, it never reads `From`. So the `fromNamespace` we pass here ("target-ns") has no meaning for this test and can mislead the reader. I would suggest to pass "" or use a helper that only sets `To`

cbrgm: but also fine if we just leave it as is, up to you :)

cbrgm:
Overall also lgtm 👍

pkg/controller/target_controller_grants_test.go
AlexHardAtWork: "target-ns" has more information then "", I'll keep it.

pkg/controller/target_controller_grants_test.go
AlexHardAtWork: Good catch, test fixed.

pkg/controller/target_controller_grants_test.go
AlexHardAtWork: @cbrgm: Good catch, test fixed.

cbrgm (approved):

PR #724, opened 2026-07-30
fix: rstructured renderartifact and -binding to solve the bug

### LINKED ISSUE #616: RenderArtifact can have stale secrets
A RenderArtifact holds a secret for the render registry to delete the OCI tag of the related rendered artifact once the RenderArtifact is deleted.

There is only one RenderArtifact for a shared rendered artifact.

Example Scenario that fails:

Target 1 gets Release A bound with a ReleaseBinding. Since Release A hasn't been bound yet to any Target, the Target Reconciler creates a RenderArtifact and make it hold the Secret reference of the render Registry. The Secret comes from the Registry that is associated with Target 1.

Target 2 now gets the same Release A bound with a ReleaseBinding. There is no RenderArtifact created since one already exists for the related rendered artifact.

Let's say Target 1 and the associated Registry and the Secret gets deleted. The RenderArtifact remain since there is still Target 2 with Release A bound. But the RenderArtifact still holds the Secret reference coming from Target 1 although already gone.

If Target 2 or its ReleaseBinding gets deleted, the RenderArtifact is supposed to be deleted as well since there is no one using the related rendered artifact anymore. The RenderArtifact finalizer tries to remove the OCI tag from the render registry with the non-existent Secret. That fails. The RenderArtifact is stuck in terminating.

Think about:
- Re-pinning the Secret of the RenderArtifact every time a RenderBinding is deleted so that it always holds a Secret of an existent RenderBinding
- Moving the Secret reference to the RenderBindings
- Something else

AC:
- A RenderArtifact has always access to the right credentials to delete the OCI tag of the related rendered artifact in the render registry within the finalizer step

rebEllieous:
## What
Changed the structure of RenderArtifacts and -bindings to reference Registries instead of having a copy of the secret
Closes #616 

## Why
RenderArtifacts had possibly stale secrets with the old implementation, and secrets would need to be repinned, dependent on remaining ArtifatBindings etc., which was convoluted and not ideal for the simple reason that we don't want the registry secrets or other sensitive data copied in too many ressources.

## Testing
make test

## Notes for reviewers
RenderArtifact lost some Registry-Specific Fields, and RenderArtifact as well as ArtifatBindings got a RegistryRef field.

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Render artifacts now reference a Registry for credentials and transport settings.
  * Render bindings preserve Registry references and support automatic artifact re-pinning.
  * Cross-namespace Registry access is supported through explicit authorization grants.
  * Registry changes are propagated to existing artifacts and bindings.

* **Bug Fixes**
  * Improved OCI cleanup authentication, host validation, and insecure-transport handling.
  * Unauthorized cleanup now reports a clear authentication failure status.

* **Documentation**
  * Updated API references, user guidance, and configuration examples for Registry references and access grants.


api/solar/v1alpha1/renderbinding_types.go
```diff
@@ -21,6 +21,16 @@ type RenderBindingSpec struct {
 	// OwnerNamespace is the namespace of the consuming resource.
 	// +kubebuilder:validation:MinLength=1
 	OwnerNamespace string `json:"ownerNamespace"`
+	// RegistryRef references the Registry this binding's owner currently resolves for
+	// pushing the shared RenderArtifact. The RenderArtifact controller re-pins
+	// RenderArtifact.Spec.RegistryRef from a surviving RenderBinding's value whenever a
+	// binding is removed, so the artifact always resolves credentials through a Registry
+	// belonging to a consumer that still exists. RenderArtifact/RenderBinding never store
+	// Secret-identifying information directly — only a reference to the Registry that owns
```
cbrgm: Inconsistent with the internal type at least my gopls linter complains :D : `api/solar/renderbinding_types.go:29` uses a comma here, this one an em-dash. Align them.

```suggestion
	// Secret-identifying information directly, only a reference to the Registry that owns
```

pkg/controller/renderartifact_controller.go
```diff
@@ -238,40 +247,129 @@ func (r *RenderArtifactReconciler) cleanupOCIArtifact(ctx context.Context, artif
...
+// among the rest the lowest name wins, so repeated reconciles converge instead of flapping
+// between equally-valid choices.
+// Because this runs on every RenderBinding create/update/delete event (see
+// mapRenderBindingToArtifact), the artifact's pinned RegistryRef is always synced to a
+// binding that exists, including immediately after the second-to-last binding is
+// removed, which is exactly the moment that matters: it leaves the artifact holding a
+// Registry reference that was valid for the binding that survives until the final
+// removal, which is what the finalizer step needs to delete the OCI tag.
+func (r *RenderArtifactReconciler) repinCredentials(ctx context.Context, artifact *solarv1alpha1.RenderArtifact, bindings []solarv1alpha1.RenderBinding) error {
+	sort.Slice(bindings, func(i, j int) bool { return bindings[i].Name < bindings[j].Name })
```
cbrgm: Prefer `slices.SortFunc` over `sort.Slice` (that's the modern go way :D), which also drops the new `sort` import since `slices`/`strings` are already imported.

```suggestion
	slices.SortFunc(bindings, func(a, b solarv1alpha1.RenderBinding) int { return strings.Compare(a.Name, b.Name) })
```

docs/superpowers/plans/2026-07-29-renderartifact-credential-repinning.md
```diff
@@ -0,0 +1,1106 @@
+# RenderArtifact Credential Re-Pinning Implementation Plan
```
cbrgm: This superpowers plan doc looks accidentally committed. Probably drop it from the PR (1106 lines of internal planning).

cbrgm:
Otherwise LGTM

cbrgm:
I unresolved the conversations which were marked as resolved since I'm pretty sure we don't want to commit the superpower stocks and I would also suggest adding the sort func at least

rebEllieous:
> I unresolved the conversations which were marked as resolved since I'm pretty sure we don't want to commit the superpower stocks and I would also suggest adding the sort func at least

we had a tiny race condition it seems :) your suggestions were/are applied

cbrgm (approved):
lgtm

PR #730, opened 2026-08-03
fix: improve handling of terminating renderartifacts

### LINKED ISSUE #619: RenderArtifact in state terminating must not fail Target reconcilation
RenderArtifact can remain in state terminating when deleting the OCI tag keeps failing.

https://github.com/opendefensecloud/solution-arsenal/blob/6a08235eeddd80683b1638689dc50794e5ead2da/pkg/controller/renderartifact_controller.go#L88-L93

That results in failing the Target reconcilation for Targets that share the related rendered artifact.

https://github.com/opendefensecloud/solution-arsenal/blob/6a08235eeddd80683b1638689dc50794e5ead2da/pkg/controller/target_controller.go#L868-L872

The name of the RenderedArtifact is deterministic as of now.

olzemal:
## What

Closes #619

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented cleanup of render artifacts that are still referenced by active bindings.
  * Improved deletion handling when artifact references are temporarily missing from the cache.
  * Deferred binding creation while an artifact is being deleted, avoiding unnecessary reconciliation errors.
  * Terminating artifacts are now preserved until dependent bindings are removed.

* **Tests**
  * Added coverage for artifact deletion, failed cleanup, retained references, deferred binding creation, and artifact creation scenarios.

### DIFF
--- a/pkg/controller/renderartifact_controller.go
+++ b/pkg/controller/renderartifact_controller.go
@@ -87,14 +87,34 @@ func (r *RenderArtifactReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// Handle deletion: attempt OCI tag cleanup, surface errors explicitly, then remove finalizer.
 	if !artifact.DeletionTimestamp.IsZero() {
 		if slices.Contains(artifact.Finalizers, renderArtifactFinalizer) {
-			if err := r.cleanupOCIArtifact(ctx, artifact); err != nil {
-				// Failure is already logged + event fired inside cleanupOCIArtifact.
-				// Keep the finalizer by returning the error so the object stays visible
-				// with the OCICleanup=False condition set.
-				return ctrl.Result{}, err
+			bound, err := r.renderArtifactBound(ctx, artifact)
+			if err != nil {
+				return ctrl.Result{}, errLogAndWrap(log, err, "failed to re-check RenderBindings for terminating RenderArtifact")
 			}
 
-			// OCI cleanup succeeded — remove finalizer to allow K8s deletion.
+			if bound {
+				// A binding still references this artifact: someone still needs the OCI tag,
+				// so keep it. The finalizer is retained on purpose: as long as the binding
+				// exists the object must not be deleted, or the tag would be orphaned in the
+				// registry (e.g. during namespace teardown nothing would recreate it). Once
+				// the binding is gone (Target deletion removes its bindings), a follow-up
+				// reconcile takes the not-bound path below and cleans up the tag.
+				log.V(1).Info("RenderArtifact is terminating but still referenced by a RenderBinding; keeping OCI tag",
+					"artifact", artifact.Name)
+				r.Recorder.Eventf(artifact, nil, corev1.EventTypeNormal, "OCICleanupSkipped", "Delete",
+					"RenderArtifact is terminating but still referenced by a RenderBinding; keeping OCI tag")
+
+				return ctrl.Result{}, nil
+			} else {
+				if err := r.cleanupOCIArtifact(ctx, artifact); err != nil {
+					// Failure is already logged + event fired inside cleanupOCIArtifact.
+					// Keep the finalizer by returning the error so the object stays visible
+					// with the OCICleanup=False condition set.
+					return ctrl.Result{}, err
+				}
+			}
+
+			// Remove finalizer to allow K8s deletion.
 			latest := artifact.DeepCopy()
 			latest.Finalizers = slices.DeleteFunc(latest.Finalizers, func(s string) bool {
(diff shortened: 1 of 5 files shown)


pkg/controller/target_controller.go
```diff
@@ -890,12 +890,19 @@ func (r *TargetReconciler) deleteOwnedRenderBindings(ctx context.Context, target
...
-			// The artifact is terminating (OCI cleanup in progress). Creating a binding
-			// against it would race with the finalizer. Requeue and wait for full deletion.
-			return fmt.Errorf("RenderArtifact %s/%s is terminating; requeuing", rt.Namespace, name)
+			// The artifact is terminating (stuck in OCI cleanup). The RenderBinding pointing
+			// at it has already been ensured, and a fresh artifact is created on a later
+			// reconcile once deletion completes. Ignore it so a stuck finalizer cannot block
+			// this Target from reconciling successfully.
+			log.V(1).Info("RenderArtifact is terminating; ignoring", "renderArtifact", key)
+
+			return nil
```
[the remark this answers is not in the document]
olzemal: @coderabbitai please check this again

cbrgm (commented):
Just minor stuff, nothing blocking.

A couple things that dont fit a clean inline suggestion:

- the three new test files use `context.Background()` throughout (14 spots at least my gopls tells me 😄 ). Since go 1.24 `t.Context()` is the idiom, it cancels with the test automatically -> worth swapping everywhere `t` is in scope
- the `RequeueAfter: 30 * time.Second` deferral shows up in both spots in `target_controller.go`. Maybe pull it into a named const so the two stay in sync if you ever tune it

pkg/controller/renderartifact_controller.go
```diff
@@ -87,14 +87,34 @@ func (r *RenderArtifactReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+				return ctrl.Result{}, nil
+			} else {
+				if err := r.cleanupOCIArtifact(ctx, artifact); err != nil {
+					// Failure is already logged + event fired inside cleanupOCIArtifact.
+					// Keep the finalizer by returning the error so the object stays visible
+					// with the OCICleanup=False condition set.
+					return ctrl.Result{}, err
+				}
+			}
```
cbrgm: nit: the `if bound` branch returns, so the `else` is dead weight. drop it and outdent the cleanup -> one less level of nesting

```suggestion
			}

			if err := r.cleanupOCIArtifact(ctx, artifact); err != nil {
				// Failure is already logged + event fired inside cleanupOCIArtifact.
				// Keep the finalizer by returning the error so the object stays visible
				// with the OCICleanup=False condition set.
				return ctrl.Result{}, err
			}
```

pkg/controller/renderartifact_controller.go
```diff
@@ -167,6 +187,34 @@ func (r *RenderArtifactReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	confirmed := &solarv1alpha1.RenderBindingList{}
+	if err := r.APIReader.List(ctx, confirmed, client.InNamespace(artifact.Namespace)); err != nil {
+		return false, err
+	}
+	for i := range confirmed.Items {
+		if confirmed.Items[i].Spec.RenderArtifactRef.Name == artifact.Name {
+			return true, nil
+		}
+	}
```
cbrgm: `slices.ContainsFunc` reads cleaner than the index loop here

```suggestion
	if slices.ContainsFunc(confirmed.Items, func(b solarv1alpha1.RenderBinding) bool {
		return b.Spec.RenderArtifactRef.Name == artifact.Name
	}) {
		return true, nil
	}
```

pkg/controller/renderartifact_finalizer_test.go
```diff
@@ -0,0 +1,183 @@
...
+		t.Helper()
+
+		got := &solarv1alpha1.RenderArtifact{}
+		if err := c.Get(context.Background(), client.ObjectKey{Name: "art", Namespace: "ns"}, got); err != nil {
+			t.Fatalf("artifact should still exist: %v", err)
+		}
+		has := slices.Contains(got.Finalizers, renderArtifactFinalizer)
+		if !has {
+			t.Error("expected the finalizer to still be present")
+		}
```
cbrgm: `has` is only used once, inline it into the `if`

```suggestion
		if !slices.Contains(got.Finalizers, renderArtifactFinalizer) {
			t.Error("expected the finalizer to still be present")
		}
```

cbrgm:
Otherwise lgtm!

cbrgm (approved):

PR #743, opened 2026-08-12
feat: deploy solar-ui as part of the solar chart

rebEllieous:
## What
Builds and publishes the `solar-ui` image in CI, and adds a `ui` component to the solar chart so it can actually be deployed.

Relates to (odd-internal) # 62

## Why
The UI is part of SolAr but had no delivery path: the image was never published, and no chart could run it

## Testing
- Built the `ui` target and ran the image: `GET /` returns the embedded `index.html` (200), `/assets/*.js` the 584 KB bündle
- `helm lint` clean. `helm template` renders the UI objects with GHCR repo, merged pull secrets and env-injected client secret, and fails as intended when `ui.enabled=true` without an issuer. Default render is identical to before

## Notes for reviewers
- **RBAC:** new ClusterRole for the UI ServiceAccount — `get`/`list` on namespaces (the BFF enumerates with its own identity, then filters per user via SelfSubjectRulesReview), plus `impersonate` on users/groups only when `ui.args.authMode=impersonate`
- **Security default:** `ui.enabled=false`, and the template `required`s `ui.oidc.issuer` / `ui.oidc.redirectURL`. Without an issuer the BFF serves the K8s API as its own ServiceAccount to anyone who reaches the pod

## Checklist
- [x] Tests added/updated
- [x] No breaking changes


charts/solar/values.yaml
```diff
@@ -323,6 +323,129 @@ controller:
...
+    redirectURL: ""
+    # -- Name of an existing Secret holding the OIDC client secret. Left empty
+    # for public clients (no client secret).
+    existingSecret: ""
+    # -- Key within `existingSecret` holding the client secret
+    existingSecretKey: "client-secret"
+
+  # UI command-line arguments
+  args:
+    # -- How the user's OIDC identity reaches the K8s API: "token" forwards the
```
cbrgm: Nit: the chart docs use "Kubernetes" everywhere else (3:1 in the README), so "K8s API" reads inconsistent. This one also flows into the generated README row.

```suggestion
    # -- How the user's OIDC identity reaches the Kubernetes API: "token" forwards the
```

charts/solar/values.yaml
```diff
@@ -323,6 +323,129 @@ controller:
...
+    pullPolicy: IfNotPresent
+
+  # -- Image pull secrets for UI
+  imagePullSecrets: []
+
+  # -- Command to run in the container
+  command: ["/solar-ui"]
+
+  # OIDC configuration. The BFF is an OIDC client; users log in against the
+  # issuer and the BFF forwards their identity to the K8s API.
```
cbrgm: Same nit here -> "Kubernetes API" for consistency.

```suggestion
  # issuer and the BFF forwards their identity to the Kubernetes API.
```

cbrgm (commented):
Only nitpicks e.g. two tiny things beyond the inline suggestions

- `ui.args.authMode` is the only key under `args`, while `extraArgs` sits right next to it as a sibling. Reads a bit odd, you'd maybe expect `ui.authMode` directly, or `extraArgs` nested under `args`. Pure naming, not important.
- `ui.imagePullSecrets` doc is terser than the siblings (renderer spells out the `[{name: my-secret}]` shape). Could mirror that one-liner so the value docs stay uniform, up to you 🤷

cbrgm:
Otherwise LGTM

rebEllieous:
> Only nitpicks e.g. two tiny things beyond the inline suggestions
> 
>     * `ui.args.authMode` is the only key under `args`, while `extraArgs` sits right next to it as a sibling. Reads a bit odd, you'd maybe expect `ui.authMode` directly, or `extraArgs` nested under `args`. Pure naming, not important.
> 
>     * `ui.imagePullSecrets` doc is terser than the siblings (renderer spells out the `[{name: my-secret}]` shape). Could mirror that one-liner so the value docs stay uniform, up to you 🤷

good point, i looked into it: both these point would make the ui section more inconsistent with the rest of the file: 
the args, exraArgs split is convention, applied that way in other sections of the chart as well. This also mirrors the python *args, **kwargs pattern (which is why its often used elsewhere) so at least to me not splitting it would read a bit odd to me 😅 
as for having it under ui.authMode: the args block, even if it only holds a single item (for now) is what's passed onto the cli call as args for the bff. THe block serves a purpose in that way, i'd prefer to keep it as is.

as for the pullSecrets: renderer is actually the outlier, ui is consistent with apiserver and controller. Renderer looks slightly different because the pod runs in the rt's namespace, so the secret has to exist in multiple namespaces

cbrgm (approved):

olzemal (approved):

PR #745, opened 2026-08-13
feat(controller): own Component lifecycle in a per-Component reconciler

### LINKED ISSUE #646: Own Component lifecycle in a per-Component reconciler (remove inferred GC + finalizer races)
## Background

#642 fixed a bug where a `Component` was orphaned when its last `ComponentVersion` (CV) was deleted (e2e webhook GC spec timing out). That fix is a deliberate band-aid (`FIXME` in `pkg/discovery/apiwriter/apiwriter.go`): it makes the apiwriter's inline parent-cleanup timing-independent, but leaves the underlying structure in place.

Two distinct concerns are currently entangled across two components:
- **Deletion protection** (the purpose of `#621`): a `Component` must not be deleted while CVs reference it. Enforced by `componentRefFinalizer`, added/removed on the `Component` by the **CV controller** (`componentversion_controller.go`).
- **GC trigger**: when the *last* CV is removed, the now-unreferenced `Component` should be deleted. Currently *inferred inline* by the **apiwriter**.

The protection is correct and must stay. The two problems are with the **trigger**:

1. **Latent re-creation race.** CV reconciles are serialized per-CV, not per-Component, and the apiwriter's trigger is an out-of-band inference. So "count CVs → act on the Component" is not atomic. If discovery re-creates a CV while another is being deleted, the Component can flap or be orphaned. `#642` does not cover this — it only removes the terminating-CV timing dependency for the delete-and-stop case.
2. **Duplicated/split logic.** Component finalize-and-delete behavior is split across two writers — the CV controller (manages the protection finalizer) and the apiwriter (issues the delete) — with no coordination between them.

## Goal

Make a single, serialized **per-Component** owner responsible for both keeping the protection finalizer in sync and triggering GC — **without changing the deletion-protection behavior** introduced in #621.

## Proposed solution

Add a `ComponentReconciler` (`pkg/controller/component_controller.go`) that:
- `For(&Component{})`, `Watches(&ComponentVersion{}, …)` mapping each CV to its parent `Component` (`spec.componentRef.name` + namespace).
- On reconcile, lists CVs via the existing `indexCVByComponentName` index and counts only **live** CVs (excludes non-zero `DeletionTimestamp`):
  - live **> 0** → **ensure `componentRefFinalizer` is present** (deletion protection — identical user-visible behavior to #621, now owned in one race-free place).
  - live **== 0** → remove `componentRefFinalizer`, then delete the `Component` (the GC trigger).

controller-runtime serializes reconciles per `Component`, so the count→act decision is atomic; a re-created CV simply re-enqueues the parent for re-evaluation. This **closes the re-creation race by construction** and makes the apiwriter's inferred trigger redundant.

### What changes
- Move `componentRefFinalizer` management from the CV controller into the `ComponentReconciler` (de-duplicate: one owner of Component finalize+delete).
- Remove the apiwriter's inline Component-delete (the #642 `FIXME` block); the writer only creates/deletes CVs.

### Explicitly preserved / out of scope
- **Deletion protection is preserved** — manual or programmatic deletion of a `Component` is still blocked while live CVs exist.
- The CV self-finalizer (`componentVersionFinalizer`) and the broader protection web (`releaseRefFinalizer`, `releaseBindingFinalizer`, `profileFinalizer` on Release/ReleaseBinding/Profile) are **untouched**. Whether the CV self-finalizer is still needed once protection is reconciler-owned is a separate evaluation, not part of this change.

## Acceptance criteria
- [ ] `ComponentReconciler` added, registered, watches CVs → enqueues parent; serialized per `Component`.
- [ ] **Protection unchanged**: `kubectl delete component X` while live CVs exist is blocked until the last CV is gone (test).
- [ ] **GC**: last CV removed → `Component` deleted; the existing e2e webhook GC spec passes (run repeatedly).
- [ ] **Re-creation safe**: deleting a CV while another CV for the same `Component` is concurrently (re)created does not orphan or flap the `Component` (new envtest).
- [ ] **De-dup**: apiwriter no longer deletes `Component`s; `componentRefFinalizer` is managed only by the `ComponentReconciler`.
- [ ] RBAC wired (`components: get;list;watch;update;delete`, `componentversions: get;list;watch`); `make manifests` + `make codegen` clean.
- [ ] Tests use envtest (real watch/index/Terminating semantics), not the fake clientset.

## Non-goals
- Adding `Release`/`ReleaseBinding`-based protection for `Component`s (never existed; out of scope).
- Changing the Release/ReleaseBinding/Profile finalizer behavior.

## References
- Band-aid fix: #642
- Deletion-protection feature: #621

cbrgm:
## What

- adds a `ComponentReconciler` (`pkg/controller/component_controller.go`) that owns the Component lifecycle: keeps `componentRefFinalizer` in sync with live ComponentVersions and garbage-collects the Component once the last live CV is gone

## Why

Component finalize-and-delete is currently split across two writers: the CV controller manages the protection finalizer, the apiwriter infers the GC trigger out-of-band -> "count CVs, act on the Component" is not atomic and re-creation races can orphan or flap the Component. A per-Component reconciler serializes the count-then-act decision per object and gives the lifecycle a single owner. Full background in the issue.

Deletion protection from #621 stays user-visible identical: `kubectl delete component X` is still blocked while live CVs exist.

Closes #646

## Testing

Whole controller suite (envtest, incl. the 5 new specs for protection, GC, re-creation race, no-CV preservation):

```
$ make test testargs="pkg/controller"
Ran 140 of 140 Specs in 141.098 seconds
SUCCESS! -- 140 Passed | 0 Failed | 0 Pending | 0 Skipped
```

Focused new specs re-ran clean multiple times, `go build ./... && go vet ./...` clean, `make manifests` regenerated `charts/solar/files/role.yaml` (components gains `delete`).

## Checklist

- [x] Tests added/updated (5 envtest specs for the new reconciler)
- [x] No breaking changes

## Summary by CodeRabbit

- **New Features**
  - Added automatic cleanup of unreferenced Components.
  - Components remain protected while active ComponentVersions reference them.
  - Added support for watching Component and ComponentVersion changes.

- **Bug Fixes**
  - Deleting a ComponentVersion no longer removes its parent Component.
  - Improved handling of concurrent deletion and recreation scenarios.
  - Preserved manually created, unreferenced Components.

### DIFF
--- a/pkg/controller/component_controller.go
+++ b/pkg/controller/component_controller.go
@@ -1,0 +1,239 @@
+// Copyright 2026 BWI GmbH and Solution Arsenal contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"context"
+	"slices"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	solarv1alpha1 "go.opendefense.cloud/solar/api/solar/v1alpha1"
+)
+
+// ComponentReconciler owns the Component lifecycle: it keeps the
+// componentRefFinalizer in sync with live ComponentVersions (deletion
+// protection) and garbage-collects a Component once its last live
+// ComponentVersion is gone. Reconciles are serialized per Component, so the
+// count-then-act decision cannot interleave with itself for the same object.
+type ComponentReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+	// APIReader reads directly from the API server, bypassing the informer
+	// cache. Used as the authoritative zero-live-CV check before deleting,
+	// because the cached index can lag behind a just-created CV.
+	APIReader client.Reader
+	// WatchNamespace restricts reconciliation to this namespace.
+	// Should be empty in production (watches all namespaces).
+	// Intended for use in integration tests only.
+	WatchNamespace string
+}
+
+//+kubebuilder:rbac:groups=solar.opendefense.cloud,resources=components,verbs=get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=solar.opendefense.cloud,resources=components/finalizers,verbs=update
+//+kubebuilder:rbac:groups=solar.opendefense.cloud,resources=componentversions,verbs=get;list;watch
+
+func (r *ComponentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	log.V(1).Info("Component is being reconciled", "req", req)
+
+	if r.WatchNamespace != "" && req.Namespace != r.WatchNamespace {
+		return ctrl.Result{}, nil
+	}
+
+	comp := &solarv1alpha1.Component{}
+	if err := r.Get(ctx, req.NamespacedName, comp); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, errLogAndWrap(log, err, "failed to get Component")
+	}
+
+	live, err := r.countLiveCVsCached(ctx, comp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if live > 0 {
+		return ctrl.Result{}, r.ensureProtectionFinalizer(ctx, comp)
+	}
+
+	// live == 0: only GC Components that carry the protection finalizer. The
+	// finalizer means a live CV was observed at some point; without it the
+	// Component was either just created by discovery (its first CV not yet
+	// visible) or created manually, and neither must be deleted here.
+	if !slices.Contains(comp.Finalizers, componentRefFinalizer) {
+		return ctrl.Result{}, nil
+	}
+
+	// Re-read the Component straight from the API server so a Component
+	// whose finalizer was already removed elsewhere (the coexisting
+	// ComponentVersionReconciler also manages componentRefFinalizer removal
+	// in this one-commit window) is left alone instead of attempting to
+	// delete it here too. This does not by itself close the delete-then-
+	// recreate race further down: the authoritative live-CV check below,
+	// and its post-delete repeat, are what actually protect a Component
+	// against a CV that shows up while GC is in flight.
+	fresh := &solarv1alpha1.Component{}
+	if err := r.APIReader.Get(ctx, req.NamespacedName, fresh); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, errLogAndWrap(log, err, "failed to get Component from API server")
+	}
+	if !slices.Contains(fresh.Finalizers, componentRefFinalizer) {
+		return ctrl.Result{}, nil
+	}
+	comp = fresh
+
+	// Authoritative re-check straight from the API server; custom field
+	// indexes are cache-only, so filter in code.
+	liveDirect, err := r.countLiveCVsDirect(ctx, comp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if liveDirect > 0 {
+		// Cache lagged behind a just-created CV; its watch event re-enqueues us.
+		return ctrl.Result{}, nil
+	}
+
+	// Delete first (the finalizer holds the object in Terminating), then strip
+	// the finalizer. If we crash in between, a later reconcile finds
+	// live == 0 with the finalizer present and finishes the removal.
+	if comp.DeletionTimestamp.IsZero() {
+		if err := client.IgnoreNotFound(r.Delete(ctx, comp, client.Preconditions{UID: &comp.UID})); err != nil {
+			return ctrl.Result{}, errLogAndWrap(log, err, "failed to delete unreferenced Component")
+		}
+		log.V(1).Info("Deleted unreferenced Component", "component", comp.Name)
+	}
+
+	// Re-check once more before stripping the finalizer: a CV can be created
+	// in the window between the delete call above and this point. If one now
+	// exists, leave the finalizer in place — the Component sits Terminating-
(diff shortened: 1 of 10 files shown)


pkg/controller/component_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	// parent (a brief hard delete under a live reference): that residual is
+	// inherent to poll-then-act without cross-resource transactions, lasts
+	// milliseconds, and self-heals because the next discovery event re-creates
+	// the Component via the apiwriter's ensureComponent.
+	liveAfterDelete, err := r.countLiveCVsDirect(ctx, comp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if liveAfterDelete > 0 {
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
cbrgm: This is the accepted residual of the delete-then-strip ordering, not an oversight. The window sits behind two authoritative live-list checks (milliseconds), needs discovery to push a brand-new CV at exactly that instant, and the state self-resolves: the Component completes deletion once that CV is gone, and the next discovery event recreates it via ensureComponent. Recreating from inside the reconciler would mean owning spec state that belongs to discovery, and k8s has no cross-resource transaction for the coordinated variant -> Terminating-and-protected is the least-bad degradation here. Documented in the comment right above the re-check; context in #746. Leaving as is.

pkg/controller/component_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	}
```
dermorz: Small thing on top of the resolved thread — I think this check makes the failure mode worse, not better.

The comment above it says: without protection, a CV created in the wrong moment loses its parent for a few milliseconds, then discovery recreates it. That's fine and self-healing.

But that's what happens if this check isn't here. What this check actually does when it fires: the Component is already Terminating at that point, and we just return. `ensureProtectionFinalizer` skips terminating objects (line 165), so no later CV event can undo it. The Component now stays Terminating until every one of its CVs is deleted. That's not milliseconds, and it doesn't heal on its own.

So we're doing a second full uncached list over the namespace to trade a self-healing problem for one that isn't. Dropping 120–134 makes the GC path half as expensive, removes ~15 lines, and leaves only the residual you already documented and accepted.

If you'd still rather have the Terminating state, let's state that in the comment. Right now it reads like it's describing what this code prevents, but it's kind of describing what happens without it.

pkg/controller/component_controller.go
```diff
@@ -0,0 +1,255 @@
...
+
+	latest := &solarv1alpha1.Component{}
+	if err := r.Get(ctx, req.NamespacedName, latest); err != nil {
```
dermorz: `r.Get` reads from the local cache, which always lags a bit behind the API server. We just changed this Component ourselves at line 114, so the cached copy is almost certainly the version from *before* that change.

The patch below refuses to apply against an outdated version (that's what `MergeFromWithOptimisticLock` does). So we get a conflict, requeue for a second, and redo everything — including both full CV lists.

`r.APIReader.Get` asks the API server directly instead of the cache, so we get the current version and the patch goes through the first time:

~~~go
if err := r.APIReader.Get(ctx, req.NamespacedName, latest); err != nil {
~~~

Probably the same reason you already use the APIReader at line 87.

pkg/controller/component_controller.go
```diff
@@ -0,0 +1,255 @@
...
+// already present or the Component is terminating (the API server rejects
+// adding finalizers to terminating objects).
+func (r *ComponentReconciler) ensureProtectionFinalizer(ctx context.Context, comp *solarv1alpha1.Component) error {
+	if !comp.DeletionTimestamp.IsZero() || slices.Contains(comp.Finalizers, componentRefFinalizer) {
+		return nil
+	}
+
+	original := comp.DeepCopy()
+	comp.Finalizers = append(comp.Finalizers, componentRefFinalizer)
+	if err := r.Patch(ctx, comp, client.MergeFrom(original)); err != nil {
```
dermorz: Minor: The list we send here comes from `comp`, read from the cache at line 53 — so if a finalizer was added since then, we drop it.

The removal path (L146) already handles this with `MergeFromWithOptimisticLock`, which makes the server reject the patch if the object changed:

~~~go
if err := r.Patch(ctx, comp, client.MergeFromWithOptions(original, client.MergeFromWithOptimisticLock{})); err != nil {
~~~

Nothing else writes Component finalizers today, so this is mostly about both paths doing the same thing.

dermorz (changes requested):
I left a few inline comments, all non-blocking except the docs.

The docs still describe the old behaviour:

- `docs/developer-guide/componentversion_controller.md` — the overview, the `component-ref` row in the finalizer table, and steps 1-2 of "On deletion" all describe `removeComponentRefFinalizer`, which this PR deletes.
- `docs/developer-guide/discovery_pipeline.md` — line 108 says the apiwriter deletes the parent Component when the last version goes, and the "Component version deleted from registry (with Component cascade)" sequence diagram still has `Writer->>K8s: Delete Component`.
- No `docs/developer-guide/component_controller.md`, and no entry in `docs/.nav.yml` — every other controller has one, and the nav is explicit so a new page needs adding by hand.

One small nit Klaus made me aware of:

- `pkg/controller/component_controller_test.go` — the "blocks Component deletion..." spec is the only one in the file without `DeferCleanup(cleanup, ...)` for `comp` and `cv`. If it ever fails partway through, the namespace teardown hangs on the finalizer instead of just reporting the failure.

pkg/controller/component_controller.go
cbrgm: Yup, added!

pkg/controller/component_controller.go
cbrgm: Also added!

pkg/controller/component_controller.go
cbrgm: Dropped the whole block in 0366a34 and rewrote the comment above the delete to state the accepted residual instead. GC path is one uncached list cheaper now, too. Funny enough i defended exactly this state on the coderabbit thread above.. consider me corrected :D

pkg/controller/component_controller.go
cbrgm: Fixed in f56ccc9 -> APIReader.Get before the strip, same reasoning as the read at the top of the GC path

pkg/controller/component_controller.go
cbrgm: Fixed in 4dbf867, add path now uses MergeFromWithOptimisticLock like the removal path :)

pkg/controller/component_controller.go
dermorz: If we continue, we might reach #nocodenoproblem ;)

dermorz (dismissed):
lgtm 🚀

dermorz (approved):

